PR #1, opened 2016-08-11
allow timestamp formats with commas

maplebed:
hack around go's inability to parse timestamps that use a comma to separate the fractional second portion.

Do a little cleanup of other portions of the file for style.


nelz9999:
Looks good.

(I tried to find any other occasions where there might be a comma, but it seems 8601 only offers that for DURATIONs. And RFC3339 seems to be a subset of 8601.)

PR #2, opened 2016-08-18
Add another comment on the --parser help

issackelly:
CAVEAT: I could have screwed this up royally because I don't `go` and I edited this from the github interface. 

`-p/--parser` now has a comment about the --list option.


christineyen:
lgtm!

PR #4, opened 2016-08-24
when called without required parameters, be helpful

maplebed:
and suggest --help instead of just 'parser required'


avifreedman:
Looks good to me...  Might be a bit more less confusing / pedantically correct to add " [optional arguments]"  after " -d <mydata>".

PR #5, opened 2016-08-30
Mysql queryparsing

seiflotfy:
Add support for:
* Query normalisation
* DB detection

### DIFF
--- a/parsers/mysql/mysql.go
+++ b/parsers/mysql/mysql.go
@@ -2,6 +2,7 @@
 package mysql
 
 import (
+	"database/sql"
 	"fmt"
 	"log"
 	"regexp"
@@ -10,11 +11,10 @@ import (
 	"sync"
 	"time"
 
-	"database/sql"
+	"github.com/honeycombio/honeytail/event"
 
 	"github.com/Sirupsen/logrus"
 	_ "github.com/go-sql-driver/mysql"
-	"github.com/honeycombio/honeytail/event"
 	"github.com/percona/go-mysql/query"
 )
 
@@ -58,9 +58,10 @@ var (
 const timeFormat = "2006-01-02T15:04:05.000000"
 
 type Options struct {
-	Host string `long:"host" description:"MySQL host in the format (address:port)"`
-	User string `long:"user" description:"MySQL username"`
-	Pass string `long:"pass" description:"MySQL password"`
+	Host          string `long:"host" description:"MySQL host in the format (address:port)"`
+	User          string `long:"user" description:"MySQL username"`
+	Pass          string `long:"pass" description:"MySQL password"`
+	QueryInterval uint   `long:"interval" description:"interval for querying the MySQL DB in seconds" default:"30"`
 }
 
 type Parser struct {
@@ -68,6 +69,7 @@ type Parser struct {
 	wg       sync.WaitGroup
 	nower    Nower
 	hostedOn string
+	readOnly bool
 }
 
 type Nower interface {
@@ -99,27 +101,53 @@ type SlowQuery struct {
 	NormalizedQuery string    `json:"normalized_query,omitempty"`
 	DB              string    `json:"db,omitempty"`
 	HostedOn        string    `json:"hosted_on,omitempty"`
+	ReadOnly        bool      `json:"read_only,omitempty"`
 	skipQuery       bool
 }
 
 func (p *Parser) Init(options interface{}) error {
 	p.conf = *options.(*Options)
 	p.nower = &RealNower{}
-	if len(p.conf.Host) > 0 {
-		p.hostedOn = getHostedOn(p.conf.Host, p.conf.User, p.conf.Pass)
+	if p.conf.Host != "" {
+		url := fmt.Sprintf("%s:%s@tcp(%s)/", p.conf.User, p.conf.Pass, p.conf.Host)
+		db, err := sql.Open("mysql", url)
+		if err != nil {
+			log.Fatal(err)
+		}
+		// update hostedOn and readOnly every <n> seconds
+		go func() {
+			defer db.Close()
+			ticker := time.NewTicker(time.Second * time.Duration(p.conf.QueryInterval))
+			for _ = range ticker.C {
+				p.hostedOn = getHostedOn(db)
+				p.readOnly = getReadOnly(db)
+			}
+		}()
 	}
 	return nil
 }
 
-func getHostedOn(host, user, pass string) string {
-	url := fmt.Sprintf("%s:%s@tcp(%s)/", user, pass, host)
-	db, err := sql.Open("mysql", url)
+func getReadOnly(db *sql.DB) bool {
+	rows, err := db.Query("SELECT @@global.read_only;")
 	if err != nil {
 		log.Fatal(err)
 	}
-	defer db.Close()
+	defer rows.Close()
 
-	rows, err := db.Query("SHOW GLOBAL VARIABLES like 'basedir';")
+	var value bool
+
+	rows.Next()
+	if err := rows.Scan(&value); err != nil {
+		log.Fatal(err)
+	}
+	if err := rows.Err(); err != nil {
+		log.Fatal(err)
+	}
+	return value
+}
+
+func getHostedOn(db *sql.DB) string {
+	rows, err := db.Query("SHOW GLOBAL VARIABLES WHERE Variable_name = 'basedir';")
 	if err != nil {
 		log.Fatal(err)
 	}
@@ -180,6 +208,7 @@ func (p *Parser) handleEvents(rawEvents <-chan rawEvent, send chan<- event.Event
 	for rawE := range rawEvents {
 		sq := p.handleEvent(rawE)
 		sq.HostedOn = p.hostedOn
+		sq.ReadOnly = p.readOnly
 		ev, err := p.processSlowQuery(sq)
 		if err != nil {
 			logrus.WithFields(logrus.Fields{
@@ -294,5 +323,6 @@ func (s SlowQuery) mapify() map[string]interface{} {
 		"query":            s.Query,
 		"normalized_query": s.NormalizedQuery,
 		"hosted_on":        s.HostedOn,
+		"read_only":        s.ReadOnly,
 	}
 }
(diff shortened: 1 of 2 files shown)


parsers/mysql/mysql.go
```diff
@@ -2,15 +2,20 @@
...
 	"sync"
 	"time"
 
-	"github.com/Sirupsen/logrus"
+	"database/sql"
 
+	"github.com/Sirupsen/logrus"
+	_ "github.com/go-sql-driver/mysql"
 	"github.com/honeycombio/honeytail/event"
+	"github.com/percona/go-mysql/query"
```
maplebed: style nit - group the standard libs together, nonstandards together, and honeytail stuff together with 1 whitespace line between each (eg logrus, mysql should be together, honeytail/event on its own line, and database/sql with the other stdlibs)

seiflotfy: took care of it :D

parsers/mysql/mysql.go
```diff
@@ -99,26 +102,54 @@ type SlowQuery struct {
...
+		log.Fatal(err)
+	}
+	if err = rows.Err(); err != nil {
+		log.Fatal(err)
+	}
+	return value
+}
 
+func getHostedOn(db *sql.DB) string {
 	rows, err := db.Query("SHOW GLOBAL VARIABLES like 'basedir';")
```
toshok: any reason to not make this `SHOW GLOBAL VARIABLES WHERE Variable_name = 'basedir';`?

parsers/mysql/mysql.go
```diff
@@ -43,17 +48,29 @@ var (
...
+	selfStr = "self"
 )
 
 const timeFormat = "2006-01-02T15:04:05.000000"
 
 type Options struct {
+	Host          string `long:"host" description:"MySQL host in the format (address:port)"`
+	User          string `long:"user" description:"MySQL username"`
+	Pass          string `long:"pass" description:"MySQL password"`
+	QueryInterval uint   `long:"interval" description:"interval for querying the MySQL DB" default:"30"`
```
maplebed: add (in seconds) to the option description

parsers/mysql/mysql.go
```diff
@@ -81,16 +98,83 @@ type SlowQuery struct {
...
+		url := fmt.Sprintf("%s:%s@tcp(%s)/", p.conf.User, p.conf.Pass, p.conf.Host)
+		db, err := sql.Open("mysql", url)
+		if err != nil {
+			log.Fatal(err)
+		}
+		p.db = db
+	}
+
+	// update hostedOn and readOnly every <n> seconds
+	go func() {
```
maplebed: Shouldn't this goroutine only be started up if p.conf.Host is set? (i.e. move it up to after line 118)

parsers/mysql/mysql.go
```diff
@@ -81,16 +98,83 @@ type SlowQuery struct {
...
+	}
+	defer rows.Close()
+
+	var value bool
+
+	rows.Next()
+	if err := rows.Scan(&value); err != nil {
+		log.Fatal(err)
+	}
+	if err = rows.Err(); err != nil {
```
maplebed: probably want err := rows.Err() here since it's within the if ... ; err {} style.

parsers/mysql/mysql.go
```diff
@@ -81,16 +97,81 @@ type SlowQuery struct {
...
+			}
+		}()
+	}
 	return nil
 }
 
+func getReadOnly(db *sql.DB) bool {
+	rows, err := db.Query("SELECT @@global.read_only;")
+	if err != nil {
+		log.Fatal(err)
```
maplebed: I'm a little concerned about the log.Fatal()s here. Our primary goal is parsing the log file and augmenting it with some extra data. If we have some transient error talking to the DB, is it reasonable to exit honeytail? Should we only exit if we fail several times in a row? or just leave read only and/or hosted on empty until they succeed  the next time? 
I don't have a good understanding of the circumstances that might cause one of these calls (db.Query, rows.Scan, or rows.Err) to fail, so I'm not sure of the severity of this issue.

seiflotfy: log fatal is wrong from my side! good call! honeytail should not exit but
rather print the error!

parsers/mysql/mysql.go
```diff
@@ -81,16 +100,145 @@ type SlowQuery struct {
...
+	rows, err := db.Query("SHOW SLAVE STATUS")
+	if err != nil {
+		log.Println(err)
+		return -1
+	}
+	if !rows.Next() {
+		log.Println(errors.New("No slave status"))
+		return -1
+	}
+	defer rows.Close()
```
toshok: this defer should be above the previous if, I think?  line 196.

parsers/mysql/mysql.go
```diff
@@ -2,15 +2,21 @@
 package mysql
 
 import (
+	"database/sql"
+	"errors"
+	"fmt"
+	"log"
```
toshok: any reason to use `log` instead of `logrus`?

parsers/mysql/mysql.go
```diff
@@ -2,15 +2,21 @@
...
 	"regexp"
 	"strconv"
 	"strings"
 	"sync"
 	"time"
 
-	"github.com/Sirupsen/logrus"
-
 	"github.com/honeycombio/honeytail/event"
+
```
toshok: import ordering should be:

```
import (
"builtin go packages"

"non-honeytail imports"

"honeytail imports"
)
```

parsers/mysql/mysql.go
```diff
@@ -81,16 +100,145 @@ type SlowQuery struct {
...
+			}
+		}()
+	}
 	return nil
 }
 
+func getReadOnly(db *sql.DB) bool {
+	rows, err := db.Query("SELECT @@global.read_only;")
+	if err != nil {
+		log.Println(err)
```
toshok: with logrus, this could be, e.g. `logrus.WithError(err).Warnln("failed to get read-only state")`

parsers/mysql/mysql.go
```diff
@@ -81,16 +100,145 @@ type SlowQuery struct {
...
+		}
+	}
+	return -1
+}
+
+func getRole(db *sql.DB) string {
+	rows, err := db.Query("SHOW MASTER STATUS")
+	if err != nil {
+		log.Println(err)
+		return "unknown"
```
toshok: this gives me pause - the behavior for missing values/errors for getHostedOn/getReadOnly/getRole is uneven.  In the error case, we'd have ""/false/"unknown" respectively.  I wonder if all three of them should be `nil` in the error case (would obviously require some re-plumbing), and then we wouldn't include those nils in the event that we send up to honeycomb.

so instead of having "false" for the read_only column, we'd have nothing if there was an error checking it.

parsers/mysql/mysql.go
```diff
@@ -81,16 +100,145 @@ type SlowQuery struct {
...
+	values := make([]interface{}, len(columns))
+	for i := range values {
+		var v sql.RawBytes
+		values[i] = &v
+	}
+
+	err = rows.Scan(values...)
+	if err != nil {
+		log.Println(err)
+		return -1
```
toshok: same comment as https://github.com/honeycombio/honeytail/pull/5/files#r77730040 - if the value we're going to send is -1, we should drop it from the event entirely.

seiflotfy:
Fixing all of this guys 👍

parsers/mysql/mysql.go
```diff
@@ -117,7 +116,7 @@ func (p *Parser) Init(options interface{}) error {
 		url := fmt.Sprintf("%s:%s@tcp(%s)/", p.conf.User, p.conf.Pass, p.conf.Host)
 		db, err := sql.Open("mysql", url)
 		if err != nil {
-			log.Fatal(err)
+			logrus.Fatal(err)
 		}
 		// run one time queries
 		p.hostedOn = getHostedOn(db)
```
toshok: I'm not sure why I didn't suggest this before - I really should have.  Given that the log messages are all the same, we should just return (*returnval, error) for each of the get* methods, and log the messages at the call sites instead of at every error check within them.

parsers/mysql/mysql.go
```diff
@@ -218,25 +217,28 @@ func getReplicaLag(db *sql.DB) int64 {
...
 
-func getRole(db *sql.DB) string {
+func getRole(db *sql.DB) *string {
+	var res string
 	rows, err := db.Query("SHOW MASTER STATUS")
 	if err != nil {
-		log.Println(err)
-		return "unknown"
+		logrus.WithError(err).Warn("failed to get role")
+		return &res
```
toshok: return nil here

parsers/mysql/mysql.go
```diff
@@ -81,16 +98,159 @@ type SlowQuery struct {
...
 	RowsExamined    int       `json:"rows_examined"`
-	Query           string    `json:"query",omitempty`
+	Query           string    `json:"query,omitempty"`
 	NormalizedQuery string    `json:"normalized_query,omitempty"`
+	DB              string    `json:"db,omitempty"`
+	HostedOn        *string   `json:"hosted_on,omitempty"`
+	ReadOnly        *bool     `json:"read_only,omitempty"`
+	ReplicaLag      *int64    `json:"replica_lag,omitempty"`
+	Role            *string   `json:"role,omitempty"`
 	skipQuery       bool
```
christineyen: do we not have a field containing the table this query touched?

christineyen:
After talking to @charity about the types of queries we'd want to be able to run on this data, I'm finding myself looking for the following information/fields and not finding it accessible via the `SlowQuery` struct:

- table name
- client ip
- query type (select, update, insert)
- rollback?

Is there another struct outside the scope of this PR that I'm not seeing, or is it just simply not available?

seiflotfy:
I can extract the query type from the query itself but table name and maybe
even the table name
in terms of client info its usually the ip and it should be there, i must
have removed it will add it again quickly!
You will have this when you wake up again!

parsers/mysql/mysql.go
```diff
@@ -328,9 +329,12 @@ func (p *Parser) handleEvent(rawE rawEvent) SlowQuery {
 			}).Debug("readmin ping detected; skipping this event")
 			sq.skipQuery = true
 		case reUser.MatchString(line):
+			fmt.Println(line)
```
toshok: cleanup

toshok:
latest change looks good, modulo updating tests

toshok:
awesome🎉

PR #10, opened 2016-09-15
adding code to validate the writekey before even trying to parse log …

maplebed:
…files to avoid sending reams of events we know will just bounce


leash.go
```diff
@@ -298,6 +298,7 @@ func sendToLibhoney(toBeSent chan event.Event, toBeResent chan event.Event,
 		default:
 		}
 		// no events at all? chill for a sec until we get the next one
+		fmt.Println("chillin")
```
toshok: do we want to print this? :)

PR #11, opened 2016-09-15
use the scan normalizer in honeycombio/mysqltools to add normalized queries

toshok:
working on a better/slower normalizer that we can either switch to entirely or when the user requests it, but for the time being get the baseline normalizer in.


christineyen:
scanNormalizer -> normalizer?

toshok:
going to close this PR in favor of #12 which parses the query and falls back to the dumb scanner if parsing fails.

PR #12, opened 2016-09-21
add mysql query normalization

toshok:
adds `normalized_query`, `tables`, and `statement` columns to the dataset.

slow queries can span several lines, so as long we aren't matching other patterns keep building up the query until we get a ';'

### DIFF
--- a/parsers/mysql/mysql.go
+++ b/parsers/mysql/mysql.go
@@ -102,7 +102,7 @@ type SlowQuery struct {
 	Role            *string  `json:"role,omitempty"`
 	ClientIP        string   `json:"client_ip,omitempty"`
 	Statement       string   `json:"statement,omitempty"`
-	Tables          []string `json:"tables,omitempty"`
+	Tables          string   `json:"tables,omitempty"`
 	skipQuery       bool
 }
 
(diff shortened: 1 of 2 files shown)


parsers/mysql/mysql.go
```diff
@@ -307,19 +312,19 @@ func (p *Parser) handleEvents(rawEvents <-chan rawEvent, send chan<- event.Event
...
 	sq := SlowQuery{}
-	logrus.WithFields(logrus.Fields{
-		"rawE": rawE,
-	}).Debug("About to parse event")
+	query := ""
 	for _, line := range rawE.lines {
 		// parse each line and populate the SlowQuery object
 		switch {
 		case reTime.MatchString(line):
+			query = ""
```
maplebed: why do you zero out query on each match? Wouldn't it be enough that it starts out empty and gets filled only by the reQuery switch case?
aaaaand query is only ever used within the reQuery case. Skip all this and just declare it local to that block?

maplebed: ahh, because it's used to build up multi-line queries. But if you start what looks like a multi-line query, then get some other line type, you want to throw out the first part of what looked like the multi-line query?

I think a comment is in order.

parsers/mysql/mysql_test.go
```diff
@@ -96,10 +99,27 @@ var sqds = []slowQueryData{
...
+		sq: SlowQuery{
+			Timestamp:       t2,
+			UnixTime:        1459470669,
+			Query:           "SELECT * FROM orders WHERE total > 1000",
+			NormalizedQuery: "select * from orders where total > ?",
+			Tables:          "orders",
+			Statement:       "select",
 		},
 	},
 	{
```
maplebed: Do you wan to insert a multi-line query to test the multi-line query joining stuff? :D

maplebed:
LGTM

PR #13, opened 2016-09-23
add a config file to honeycomb

maplebed:
Make it possible to configure honeytail using a config file instead of command line flags. 

For a sample config file, run ./honeytail --write_default_config > ./default_honeytail_config.ini

To convert an existing honeytail command line to a config file, add --write_current_config to the command line and save the output.

### DIFF
--- a/main.go
+++ b/main.go
@@ -86,7 +86,6 @@ func main() {
 	flagParser := flag.NewParser(&options, flag.PrintErrors)
 	flagParser.Usage = "-p <parser> -k <writekey> -f </path/to/logfile> -d <mydata> [optional arguments]"
 
-	// parse flags and check for extra command line args
 	if extraArgs, err := flagParser.Parse(); err != nil || len(extraArgs) != 0 {
 		fmt.Println("Error: failed to parse the command line.")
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -80,6 +84,13 @@ func main() {
 	var options GlobalOptions
 	flagParser := flag.NewParser(&options, flag.PrintErrors)
 	flagParser.Usage = "-p <parser> -k <writekey> -f </path/to/logfile> -d <mydata> [optional arguments]"
+
+	// read the config file if present
+	options.ConfigFile = func(s string) error {
```
toshok: funky, you can change `options` after creating `flagParser`?

main.go
```diff
@@ -80,6 +84,13 @@ func main() {
...
+		ini := flag.NewIniParser(flagParser)
+		return ini.ParseFile(s)
+	}
+
 	if extraArgs, err := flagParser.Parse(); err != nil || len(extraArgs) != 0 {
```
toshok: I can't remember the decision wrt enforcing the `--config` option coming first on the command line.  Is that something we want to do?

PR #14, opened 2016-09-29
Remove SlowQuery struct in lieu of a map

christineyen:
We're building a map[string]interface{} for the `event.Event` ultimately anyway,
so we may as well use it directly in the first place. This ensures that
we omit missing values, rather than sending along zero-values in the
struct itself.

### DIFF
--- a/parsers/mysql/mysql.go
+++ b/parsers/mysql/mysql.go
@@ -41,6 +41,31 @@ import (
 // We should ignore the administrator command entry; the stats it presents (eg rows_sent)
 // are actually for the previous command
 
+const (
+	rdsStr  = "rds"
+	ec2Str  = "ec2"
+	selfStr = "self"
+
+	// Event attributes
+	userKey            = "user"
+	clientKey          = "client"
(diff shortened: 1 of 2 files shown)


parsers/mysql/mysql.go
```diff
@@ -325,38 +331,38 @@ func (p *Parser) handleEvent(rawE []string) (SlowQuery, time.Time) {
 				"line":  line,
 				"event": rawE,
 			}).Debug("readmin ping detected; skipping this event")
-			sq.skipQuery = true
+			return nil, time.Time{}
```
toshok: does `len(nil)` == 0 (line 288)?  because if so, cool.

christineyen: it does when it's a nil masquerading as a `map[string]interface{}`! one of the hilariously fun failures of `reflect.DeepEqual` vs `==`.

https://play.golang.org/p/JHTuJsDgLn

toshok: wow, love it

toshok:
🎉

PR #15, opened 2016-09-30
don't use `# Time` as the event delimiter.

toshok:
rds is only inserting a line every second, so we were skipping many, many queries (we were sending only about 1/100th of the entries.)


christineyen:
LGTM

PR #16, opened 2016-10-11
fix mysql line -> event splitting

maplebed:
Adjust logic for splitting mysql events to force finding at least one non-comment before finding a comment indicates that an event is complete

### DIFF
--- a/parsers/mysql/mysql.go
+++ b/parsers/mysql/mysql.go
@@ -267,16 +267,11 @@ func (p *Parser) ProcessLines(lines <-chan string, send chan<- event.Event) {
 	go p.handleEvents(rawEvents, send)
 
 	// flag to indicate when we've got a complete event to send
-	var sendEvent bool
 	var foundStatement bool
 	groupedLines := make([]string, 0, 5)
 	for line := range lines {
-		if foundStatement && strings.HasPrefix(line, "# ") && len(groupedLines) > 0 {
+		if foundStatement && len(groupedLines) > 0 && strings.HasPrefix(line, "# ") {
 			// we've started a new event. Send the previous one.
-			sendEvent = true
-		}
-		if sendEvent {
-			sendEvent = false
 			foundStatement = false
 			rawEvents <- groupedLines
 			groupedLines = make([]string, 0, 5)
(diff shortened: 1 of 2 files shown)


parsers/mysql/mysql.go
```diff
@@ -260,20 +267,23 @@ func (p *Parser) ProcessLines(lines <-chan string, send chan<- event.Event) {
 	go p.handleEvents(rawEvents, send)
 
 	// flag to indicate when we've got a complete event to send
-	var sendEvent bool
+	var foundStatement bool
 	groupedLines := make([]string, 0, 5)
 	for line := range lines {
-		if strings.HasPrefix(line, "# ") && len(groupedLines) > 0 {
```
toshok: gah, so we were sending events for single comment lines before?  :facepalm:

parsers/mysql/mysql.go
```diff
@@ -260,20 +267,23 @@ func (p *Parser) ProcessLines(lines <-chan string, send chan<- event.Event) {
...
+		if foundStatement && len(groupedLines) > 0 && strings.HasPrefix(line, "# ") {
```
toshok: miiiiiiild preference for a pulling the HasPrefix call out of both if's:  `lineIsComment := strings.HasPrefix(line, "# ")`.

and/or possibly hoisting the `HasPrefix` out:

```
if lineIsComment {
  /* rest of multiline check */
} else {
  /* stuff in the !HasPrefix if below */
}
```

toshok: can it ever be the case that `foundStatement == true` but `len(groupedLines) <= 0`?  It doesn't look like it.

parsers/mysql/mysql.go
```diff
@@ -260,20 +267,23 @@ func (p *Parser) ProcessLines(lines <-chan string, send chan<- event.Event) {
...
 		}
+		if !strings.HasPrefix(line, "# ") {
+			// we've finished the comments before the statement and now should slurp
+			// lines until the next comment
+			foundStatement = true
+		}
 		groupedLines = append(groupedLines, line)
 	}
+	// send the last event, if there was one collected
 	if len(groupedLines) != 0 {
```
toshok: same here, looks like this could be `if foundStatement`

parsers/mysql/mysql_test.go
```diff
@@ -309,6 +334,144 @@ func TestTimeProcessing(t *testing.T) {
 	}
 }
 
+// test that ProcessLines correctly splits the mysql slow query log stream into
```
toshok: 💯 👍

PR #17, opened 2016-10-14
Make MySQL query stat parsing a little more flexible, add a few more fields

christineyen:
Permit `Rows_affected` at the end of the query stats line and parse useful-looking values out of a series of heretofore ignored InnoDB log lines.


toshok:
https://www.percona.com/doc/percona-server/5.5/diagnostics/slow_extended_55.html#id1 is where all the percona innodb fields are described.

toshok:
we have two time fields without units in the mysql parser: `query_time` and `lock_time`.  These should both be suffixed with `_sec`, but that'll require a change to common queries.

PR #22, opened 2016-10-19
add detection of mysql headers that rds puts into the slow query log. end events when we see them, but otherwise ignore.

toshok:
according to http://dba.stackexchange.com/questions/56945/why-is-my-rds-mysql-slow-query-log-full-of-server-restarts rds inserts the headers whenever logs are flushed.

For now just ignore the lines (in the future it might be useful to track the version of mysql?) except to treat them as ending the previous event (same as we do for comment lines now.)


maplebed:
LGTM

PR #23, opened 2016-10-20
make auto-columnifying of query parameters optional (and default off)

maplebed:
The internet throws massive cruft at web servers, and the glut of useless query parameter keys is overwhelming. They pollute the dataset and make it more difficult to use. Switch to a whitelist of expected keys instead (by default) but allow the user to override that and create every key as a column.

### DIFF
--- a/leash.go
+++ b/leash.go
@@ -242,6 +242,18 @@ func requestShape(field string, toBeSent chan event.Event, options GlobalOptions
 		}
 		pr.Patterns = append(pr.Patterns, &pat)
 	}
+	// define local inclusion test so options are in scope
+	shouldIncludeQueryKey := func(key string) bool {
+		if options.RequestParseQuery == "all" {
+			return true
+		}
+		for _, whiteKey := range options.RequestQueryKeys {
+			if key == whiteKey {
+				return true
+			}
+		}
+		return false
+	}
 	go func() {
 		for ev := range toBeSent {
 			if val, ok := ev.Data[field]; ok {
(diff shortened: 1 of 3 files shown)


leash.go
```diff
@@ -268,12 +268,12 @@ func requestShape(field string, toBeSent chan event.Event, options GlobalOptions
 				ev.Data[prefix+field+"_path"] = res.Path
 				ev.Data[prefix+field+"_query"] = res.Query
 				for k, v := range res.QueryFields {
-					if len(v) == 0 {
-						ev.Data[prefix+field+"_query_"+k] = ""
-					} else if len(v) == 1 {
```
christineyen: 👍

main.go
```diff
@@ -46,13 +46,15 @@ type GlobalOptions struct {
...
+	AddFields         []string `long:"add_field" description:"Add the field to every event. Field should be key=val. May be specified multiple times"`
+	RequestShape      []string `long:"request_shape" description:"Identify a field that contains an HTTP request of the form 'METHOD /path HTTP/1.x' or just the request path. Break apart that field into subfields that contain components. May be specified multiple times. Defaults to 'request' when using the nginx parser"`
+	RequestPattern    []string `long:"request_pattern" description:"A pattern for the request path on which to base the derived request_shape. May be specified multiple times. Patterns are considered in order; first match wins."`
+	RequestParseQuery string   `long:"request_parse_query" description:"How to parse the request query parameters. 'whitelist' means only extract listed query keys. 'all' means to extract all query parameters as individual columns" default:"whitelist"`
+	RequestQueryKeys  []string `long:"request_query_keys" description:"Request query parameter key names to extract, when request_parse_query is 'whitelist'. May be specified multiple times."`
```
christineyen: let's please make sure to write out some specific, short examples of this behavior in our docs. there are enough options here to be a little overwhelming and this is a lot of text to read.

main.go
```diff
@@ -46,13 +46,15 @@ type GlobalOptions struct {
...
+	ShapePrefix       string   `long:"shape_prefix" description:"Prefix to use on fields generated from request_shape to prevent field collision"`
```
christineyen: presumably `ShapePrefix` is meant to affect `RequestShape` and `RequestPattern`? can we move it above `RequestParseQuery`/`RequestQueryKeys` so someone scanning through the list sees those grouped together?

main.go
maplebed: will do.

PR #24, opened 2016-10-24
skip over statefiles when tailing files

maplebed:
When repeatedly running honeytail with --file=/some/path/\*, you wind up getting files named foo.leash.state.leash.state.leash.state.etc.etc. That's silly and we shouldn't do it.
When tailing files, skip those that are named the same as the `--tail.statefile` flag or the default of `*.leash.state`


christineyen:
LGTM

PR #25, opened 2016-10-28
Mongo 3.4 tests, sharding support, and pseudo-queries

toshok:


parsers/mongodb/mongodb_test.go
```diff
@@ -19,6 +19,13 @@ const (
...
+	UPDATE_COMMAND      = `Tue Sep 13 21:10:33.961 I COMMAND  [conn11896572] command data.$cmd command: update { update: "avengers", updates: [ { q: { hulkForm: "Bruce Banner" }, u: { $set: { hulkForm: "Big Green" }, $setOnInsert: { hulkForm: "Big Green" } }, upsert: true } ], writeConcern: { getLastError: 1, w: 1 }, ordered: true } keyUpdates:0 writeConflicts:0 numYields:0 reslen:95 locks:{ Global: { acquireCount: { r: 1, w: 1 } }, Database: { acquireCount: { w: 1 } }, Collection: { acquireCount: { w: 1 } } } user_key_comparison_count:466 block_cache_hit_count:10 block_read_count:0 block_read_byte:0 internal_key_skipped_count:17 internal_delete_skipped_count:0 get_from_memtable_count:0 seek_on_memtable_count:2 seek_child_seek_count:12 0ms`
```
christineyen: in the interest of using the test cases as a "here's what to expect for X type of log line," think we can get simpler update / delete cases represented alongside these multi-body ones?

toshok: sure 👍

toshok:
the two new log lines correspond to `update foo set bar = 1 where baz = 1`, and `delete from foo where bar = 1`, respectively.  No complicated upsert/setOnInsert stuff.

PR #26, opened 2016-10-31
Add a POC for support for ArangoDB.

neunhoef:
This is rather limited as it stands, it can only interpret a single type of log entry, namely the logging of HTTP requests.

### DIFF
--- a/parsers/arangodb/arangodb.go
+++ b/parsers/arangodb/arangodb.go
@@ -1,0 +1,199 @@
+// Package arangodb is a parser for ArangoDB logs
+package arangodb
+
+import (
+	"errors"
+	"strings"
+	"strconv"
+	"time"
+	"fmt"
+
+	"github.com/Sirupsen/logrus"
+	"github.com/honeycombio/honeytail/event"
+)
+
+const (
+	iso8601UTCTimeFormat   = "2006-01-02T15:04:05Z"
+	iso8601LocalTimeFormat = "2006-01-02T15:04:05"
+
+	timestampFieldName   = "timestamp"
+	pidFieldName         = "pid"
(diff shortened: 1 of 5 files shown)


maplebed:
This is awesome. Thank you! I'm looking forward to reviewing ASAP.

parsers/arangodb/arangodb.go
```diff
@@ -0,0 +1,196 @@
...
+	if word[0] == '"' {
+		word = word[1:]
+  }
+	if len(word) > 0 && word[len(word)-1] == '"' {
+		word = word[:len(word)-1]
+	}
+	return word
+}
+
+func (m *ArangoLineParser) ParseLogLine(line string) (values map[string]interface{}, err error) {
```
maplebed: Since values is only used to be an empty map on errored return, let's change it to an unnamed variable `(_ map[string]interface{}, err error)`

parsers/arangodb/arangodb.go
```diff
@@ -0,0 +1,196 @@
...
+				return timestamp, nil
+			}
+		}
+		return time.Time{}, err
+	}
+
+	return time.Time{}, errors.New("timestamp missing from logline")
+}
+
+func logFailure(line string, err error, msg string) {
```
maplebed: It's intentional that only request lines are going to be parsed - all the other line types aren't errors, just omitted. Let's log them as INFO or DEBUG instead of ERROR level.

neunhoef:
Thanks for the feedback. This is the first go program I write, so any hints are very much appreciated. I have tried to address the issues. Please let me know if there is anything else.

maplebed:
Looks great!  Thanks! I'm super excited by this. :)

PR #28, opened 2016-11-08
Add periodic status lines for file tailing stats and parsing stats

maplebed:
Add in one last printout of the stats when honeytail finishes.  This is a huge help when parsing short files with --tail.stop and honeytail says "All done" and you wonder "what happened?".

Add in stats from file tailing (how many lines, how many errors, how far behind the tip of the file) and parsing (whatever the parsing module exports, but likely counting lines and events and parsing errors).

### DIFF
--- a/leash.go
+++ b/leash.go
@@ -24,6 +24,8 @@ import (
 // hny is a logfile / tailer / parser combo for concentrating stats info into
 // one place
 type hny struct {
+	parser parsers.Parser
+	tailer *tail.Tailer
 	rStats *responseStats
 }
 
@@ -55,7 +57,7 @@ func run(options GlobalOptions) {
 	}
 
 	// get our lines channel from which to read log lines
-	linesChans, err := tail.GetEntries(tail.Config{
+	tailers, err := tail.GetEntries(tail.Config{
 		Paths:   options.Reqs.LogFiles,
 		Type:    tail.RotateStyleSyslog,
 		Options: options.Tail})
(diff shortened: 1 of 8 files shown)


maplebed:
I would love suggestions on a better name for the list of honeytail threads than `hnys`

leash.go
```diff
@@ -100,23 +109,39 @@ func run(options GlobalOptions) {
...
 			parsersWG.Done()
-		}(lines)
+		}(tailer)
+		hny := &hny{
+			parser: parser,
+			tailer: tailer,
+			rStats: stats,
+		}
+		hnys = append(hnys, hny)
+		go logStats(hny, options.StatusInterval)
```
christineyen: afaict, the purpose of the `hny` struct is basically because you want to do one last pass over everything at line 139. and you only need to do that because you want to do it one last time before everything shuts down, but you need access to these locally-defined things (`parser`, `tailer`, `stats`).

idea: instead of firing off `len(tailers)` number of `logStats` goroutines, each with their own set of 3 things to log out + their own ticker + their own early return, why not group all of the things that we want to log out and do all of them at once, in a ticker loop and on shutdown?

bonus: `tailer` and `parser` already have a `LogStats()` function defined, and this is the only invocation of `logAndReset` on `responseStats`, so you can rename *that* easily.

suggestion -- define a:

    type StatsLogger interface{
      LogStats()
    }

and instead of building up a `hny{}` and then a `[]hny{}`, just build up a `[]StatsLogger` for everything, then call `go logStats` *once*, outside the `for ... range tailers` loop, and once right before shutdown.

your instinct that the `hny` struct was weird was spot on. do less work.

christineyen: also, let's sanity check that the `tailer` logging does the right thing when multiple files are specified.

christineyen:
Closing in favor of #75

PR #29, opened 2016-11-09
adding support for doing sampling while tailing

maplebed:
instead of after parsing to save CPU. Setting it as default for all parsers but MySQL.

### DIFF
--- a/main.go
+++ b/main.go
@@ -213,6 +213,10 @@ func sanityCheckOptions(options *GlobalOptions) {
 		fmt.Println("Dataset name required.")
 		usage()
 		os.Exit(1)
+	case options.SampleRate == 0:
+		fmt.Println("Sample rate must be an integer >= 1")
+		usage()
+		os.Exit(1)
 	case options.Tail.ReadFrom == "end" && options.Tail.Stop:
 		fmt.Println("Reading from the end and stopping when we get there. Zero lines to process. Ok, all done! ;)")
 		usage()
--- a/main.go
+++ b/main.go
@@ -37,7 +37,7 @@ var validParsers = []string{
 // GlobalOptions has all the top level CLI flags that honeytail supports
 type GlobalOptions struct {
 	APIHost    string `hidden:"true" long:"api_host" description:"Host for the Honeycomb API" default:"https://api.honeycomb.io/"`
-	TailSample bool   `hidden:"true" long:"tail_sample" description:"When true, sample while tailing. When false, sample post-parser events"`
+	TailSample bool   `hidden:"true" description:"When true, sample while tailing. When false, sample post-parser events"`
 
 	ConfigFile string `short:"c" long:"config" description:"Config file for honeytail in INI format." no-ini:"true"`
 
--- a/main.go
+++ b/main.go
@@ -36,7 +36,8 @@ var validParsers = []string{
 
 // GlobalOptions has all the top level CLI flags that honeytail supports
 type GlobalOptions struct {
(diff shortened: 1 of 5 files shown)


main.go
```diff
@@ -36,7 +36,8 @@ var validParsers = []string{
 
 // GlobalOptions has all the top level CLI flags that honeytail supports
 type GlobalOptions struct {
-	APIHost string `hidden:"true" long:"api_host" description:"Host for the Honeycomb API" default:"https://api.honeycomb.io/"`
+	APIHost    string `hidden:"true" long:"api_host" description:"Host for the Honeycomb API" default:"https://api.honeycomb.io/"`
+	TailSample bool   `hidden:"true" long:"tail_sample" description:"When true, sample while tailing. When false, sample post-parser events"`
```
christineyen: -> "When true, drop 'sampled' lines before handing them to the parser. Should not be used for log types likely to have multi-line events (e.g. MySQL)"

main.go
```diff
@@ -185,6 +186,13 @@ func addParserDefaultOptions(options *GlobalOptions) {
 		// automatically normalize the request when using the nginx parser
 		options.RequestShape = append(options.RequestShape, "request")
 	}
+	if options.Reqs.ParserName != "mysql" {
+		// mysql is the only parser that requires post-parsed sampling.
+		// Sample all other parser when tailing to conserve CPU
+		options.TailSample = true
+	} else {
+		options.TailSample = false
```
christineyen: won't this just clobber anything that gets passed in? is the intent for the flag to be purely a global switch, rather than an actual settable flag?

maplebed:
test will fail until libhoney with support for presampled send gets merged to master.

tail/tail.go
```diff
@@ -57,6 +58,46 @@ type State struct {
...
 }
 
+// GetSampledEntries wraps GetEntries and returns a list of channels that
+// provide sampled entries
+func GetSampledEntries(conf Config, sampleRate uint) ([]chan string, error) {
+	unsampledLinesChans, err := GetEntries(conf)
+	if err != nil {
+		return nil, err
+	}
+	if sampleRate == 1 {
```
christineyen: we're sure `sampleRate` will be `1` here and that we shouldn't be checking `<=`? (what *is* the normal boundary for "if someone tells us '0' for sample rate they really mean '1'"?

tail/tail.go
```diff
@@ -57,6 +58,46 @@ type State struct {
...
+	sampledLinesChans := make([]chan string, 0, len(unsampledLinesChans))
+
+	for _, lines := range unsampledLinesChans {
+		sampledLines := make(chan string)
+		go func(pLines chan string) {
+			defer close(sampledLines)
+			for line := range pLines {
+				if shouldDrop(sampleRate) {
+					logrus.WithFields(logrus.Fields{
+						"line": line,
```
christineyen: if you're going to log, stick the `sampleRate` in there.

the debug output is getting a little unwieldy, but that's a point for another time, i think.

main.go
maplebed: Sticking this in options, but the `hidden:true` effectively means it's not user-facing. It won't show up in --help so the only way to set it would be if you knew it existed. 
I put it in options because that seemed the most reasonable way to pass it around along with all the other bits. 
hmm.  Removing the long:"tail_sample" bit might help it be even more non-user-settable. I'll do that.

main.go
maplebed: yup. Purely a global switch.

PR #30, opened 2016-11-16
Add prefix stripping to honeytail

maplebed:
In order to facilitate log forwarders adding prefixes to the log lines, add a flag that allows you to strip those headers (potentially grabbing useful information like the hostname) and pass the original line through to the parser


parsers/mysql/mysql.go
```diff
@@ -314,6 +315,14 @@ func (p *Parser) ProcessLines(lines <-chan string, send chan<- event.Event) {
 	var foundStatement bool
 	groupedLines := make([]string, 0, 5)
 	for line := range lines {
+		// mysql parser does not support capturing fields in the line prefix - just
+		// strip it.
+		if prefixRegex != nil {
+			var prefix string
+			prefix, _ = prefixRegex.FindStringSubmatchMap(line)
```
christineyen: if you're not using the `-Map` functionality of this variant, let's just use `prefixRegex.FindString` directly.

parsers/helpers.go
```diff
@@ -0,0 +1,30 @@
+package parsers
```
christineyen: _registering mild grumpiness at "helpers.go" as a generic/unhelpful filename, but without a significantly better suggestion than "regexp_helpers.go"_

parsers/helpers.go
```diff
@@ -0,0 +1,30 @@
...
+
+import "regexp"
+
+// ExtRegexp is a Regexp with one additional method FindStringSubmatchMap
```
christineyen: s/`FindStringSubmatchMap`/ to make it easier to work with named groups/

PR #32, opened 2016-11-29
Fix leash_test to prepare for gzipped/batched events by default

christineyen:
Fixes tests once https://github.com/honeycombio/libhoney-go/pull/3 is landed, to handle batching/gzipping behavior.

(Tests will fail until ^ is merged)

### DIFF
--- a/leash_test.go
+++ b/leash_test.go
@@ -33,9 +33,10 @@ var tailOptions = tail.TailOptions{
 // defaultOptions is a fully populated GlobalOptions with good defaults to start from
 var defaultOptions = GlobalOptions{
 	// each test will have to populate APIHost with the location of its test server
-	APIHost:    "",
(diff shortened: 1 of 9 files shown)


leash_test.go
```diff
@@ -209,15 +223,15 @@ func TestDropField(t *testing.T) {
 	opts.Reqs.LogFiles = []string{logFileName}
 	run(opts)
 	testEquals(t, ts.rsp.reqCounter, 1)
-	testEquals(t, ts.rsp.reqBody, `{"dropme":"chew","format":"json","reallygone":"notyet"}`)
+	testContains(t, ts.rsp.reqBody, `{"dropme":"chew","format":"json","reallygone":"notyet"}`)
```
maplebed: lovely trick - batch or single, contains will still be true.  \o/

PR #33, opened 2016-12-20
adding PathShape and QueryShape

maplebed:
adding PathShape and QueryShape to access those halves of the entire request shape independently


maplebed:
tests will pass after https://github.com/honeycombio/urlshaper/pull/2 gets merged

leash.go
```diff
@@ -303,6 +303,8 @@ func requestShape(field string, toBeSent chan event.Event, options GlobalOptions
 					ev.Data[prefix+field+"_path_"+k] = v[0]
 				}
 				ev.Data[prefix+field+"_shape"] = res.Shape
+				ev.Data[prefix+field+"_pathshape"] = res.PathShape
+				ev.Data[prefix+field+"_queryshape"] = res.QueryShape
```
christineyen: strong preference for omitting empty string fields, especially for these sorts of "breakdown" columns (path vs query)

(also on lines 290/291: `res.Path` vs `res.Query`)

PR #41, opened 2017-02-28
[json] Warn on failure to parse times

christineyen:
Thought about adding a test and downgraded it to just a warning. This will trigger only once per run. (Also note that that URL/alias isn't live yet but should be, later tonight)

image

### DIFF
--- a/parsers/htjson/htjson.go
+++ b/parsers/htjson/htjson.go
@@ -30,6 +30,8 @@ type Parser struct {
 	conf       Options
 	lineParser LineParser
 	nower      Nower
+
+	warnedAboutTime bool
 }
 
 type Nower interface {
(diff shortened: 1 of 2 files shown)


parsers/htjson/htjson.go
```diff
@@ -203,3 +206,11 @@ func (p *Parser) tryTimeFormats(t string) time.Time {
 	}
 	return ts
 }
+
+func (p *Parser) warnAboutTime(fieldName string, foundTimeVal interface{}, msg string) {
+	if p.warnedAboutTime {
+		return
+	}
+	logrus.WithField("time_field", fieldName).WithField("time_value", foundTimeVal).Warn(msg + "\n  Please refer to https://honeycomb.io/docs/json#timestamp-parsing")
```
maplebed: <3 I was about to say "let's put the time field spec and contents in the error message" up above.  \o/ structured logging!

PR #43, opened 2017-03-14
push sampling for mysql logs into the mysql parser

maplebed:
Instead of relying on libhoney's sampling, have the mysql parser do it to avoid wasting an enormous amount of CPU parsing lines we're just going to throw away.

### DIFF
--- a/leash.go
+++ b/leash.go
@@ -113,9 +113,9 @@ func run(options GlobalOptions) {
 			close(realToBeSent)
 		}()
 
-		// start up the sender
-		go sendToLibhoney(realToBeSent, toBeResent, delaySending, doneSending,
-			options.TailSample)
+		// start up the sender. all sources are either sampled when tailing or in-
+		// parser, so always tell libhoney events are pre-sampled
+		go sendToLibhoney(realToBeSent, toBeResent, delaySending, doneSending, true)
 
 		// start a goroutine that reads from responses and logs.
 		responses := libhoney.Responses()
@@ -156,7 +156,9 @@ func getParserAndOptions(options GlobalOptions) (parsers.Parser, interface{}) {
 		parser = &mongodb.Parser{}
 		opts = &options.Mongo
 	case "mysql":
-		parser = &mysql.Parser{}
+		parser = &mysql.Parser{
+			SampleRate: int(options.SampleRate),
+		}
 		opts = &options.MySQL
 	case "arangodb":
 		parser = &arangodb.Parser{}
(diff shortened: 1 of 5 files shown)


parsers/mysql/mysql.go
```diff
@@ -331,15 +336,32 @@ func (p *Parser) ProcessLines(lines <-chan string, send chan<- event.Event, pref
 			if foundStatement {
 				// we've started a new event. Send the previous one.
 				foundStatement = false
-				rawEvents <- groupedLines
+				// if sampling is enabled, drop events here to conserve parsing CPU
+				if p.SampleRate > 1 {
+					if rand.Intn(p.SampleRate) == 0 {
```
emfree: nit: could shorten to 
```
if p.SampleRate > 1 && rand.Intn(p.SampleRate) == 0 {
```
here and below

christineyen: i think we don't want that, or else we'll hit line 347 if `p.SampleRate > 1` and we should drop it (aka `rand.Intn(p.SampleRate) != 0`)

christineyen: it COULD be flipped to read:

```
if p.SampleRate <= 1 || rand.Intn(p.SampleRate) == 0 {
  rawEvents <- groupedLines
}
```

(and ditto down below)

leash.go
```diff
@@ -113,9 +113,9 @@ func run(options GlobalOptions) {
 			close(realToBeSent)
 		}()
 
-		// start up the sender
-		go sendToLibhoney(realToBeSent, toBeResent, delaySending, doneSending,
-			options.TailSample)
+		// start up the sender. all sources are either sampled when tailing or in-
+		// parser, so always tell libhoney events are pre-sampled
+		go sendToLibhoney(realToBeSent, toBeResent, delaySending, doneSending, true)
```
christineyen: if this is the only invocation of `sendToLibHoney` and we always pass in `true`, is... it worth keeping that `presampled bool` param in place? are we likely to want to use it again in the future?

parsers/mysql/mysql.go
emfree: doh! right.

leash.go
maplebed: I was on the fence but I think "add it back if we need it again" will make future us happier than "keep around this thing that's sorta confusing and not actually used anywhere".

parsers/mysql/mysql.go
maplebed: done.

PR #44, opened 2017-03-15
add signal handling to cleanly exit on SIGINT or SIGTERM.

maplebed:
also mix up tailing a file so we don't need an additional line after receiving the signal in order to process it.

### DIFF
--- a/leash.go
+++ b/leash.go
@@ -3,10 +3,14 @@ package main
 import (
 	"crypto/sha256"
 	"fmt"
+	"io"
+	"os"
+	"os/signal"
 	"regexp"
 	"sort"
 	"strings"
 	"sync"
+	"syscall"
 	"time"
 
 	"github.com/Sirupsen/logrus"
@@ -26,6 +30,22 @@ import (
 // actually go and be leashy
 func run(options GlobalOptions) {
 	logrus.Info("Starting honeytail")
+
+	sigs := make(chan os.Signal, 1)
+	abort := make(chan bool, 1)
+	signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
+	go func() {
+		sig := <-sigs
+		io.WriteString(os.Stderr, fmt.Sprintf("Aborting! Caught signal \"%s\"\n", sig))
+		io.WriteString(os.Stderr, "Cleaning up...\n")
+		abort <- true
+		go func() {
+			// don't wait forever to clean up. if we're really stuck, just die.
+			time.Sleep(10 * time.Second)
+			io.WriteString(os.Stderr, "Taking too long to clean up. I die, I die.\n")
+			os.Exit(1)
+		}()
+	}()
 
 	// spin up our transmission to send events to Honeycomb
 	libhConfig := libhoney.Config{
@@ -67,9 +87,9 @@ func run(options GlobalOptions) {
 		Options: options.Tail,
 	}
 	if options.TailSample {
-		linesChans, err = tail.GetSampledEntries(tc, options.SampleRate)
+		linesChans, err = tail.GetSampledEntries(tc, options.SampleRate, abort)
 	} else {
-		linesChans, err = tail.GetEntries(tc)
+		linesChans, err = tail.GetEntries(tc, abort)
 	}
 	if err != nil {
 		logrus.WithFields(logrus.Fields{"err": err}).Fatal(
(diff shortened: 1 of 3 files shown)


tail/tail.go
```diff
@@ -191,12 +191,25 @@ func tailSingleFile(tailer *tail.Tail, file string, stateFile string) chan strin
...
+				if line.Err != nil {
+					// skip errored lines
+					continue
+				}
+				lines <- strings.TrimSpace(line.Text)
+			case <-abort:
+				break ReadLines
+			default:
+				// no new lines to process, just chill for a bit
+				time.Sleep(50 * time.Millisecond)
```
emfree: hm, is this sleep case necessary?

maplebed: yes - if it's a quiet log file and there are no updates we don't want to spin on trying to read it because that eats up CPU wastefully.

tail/tail.go
```diff
@@ -208,12 +218,19 @@ func tailSingleFile(tailer *tail.Tail, file string, stateFile string) chan strin
...
 	lines := make(chan string)
 	input := bufio.NewReader(os.Stdin)
 	go func() {
 		defer close(lines)
+	ReadLines:
 		for {
+			// check for signal triggered exit
+			select {
+			case <-abort:
+				break ReadLines
```
christineyen: why not just `return`?

tail/tail.go
```diff
@@ -191,12 +191,22 @@ func tailSingleFile(tailer *tail.Tail, file string, stateFile string) chan strin
...
+					// tailer.Lines is closed
+					break ReadLines
+				}
+				if line.Err != nil {
+					// skip errored lines
+					continue
+				}
+				lines <- strings.TrimSpace(line.Text)
+			case <-abort:
+				break ReadLines
```
christineyen: 🤔 how does ordering impact behavior here? seems like precedence here would let you not have to wait for a line to come in

maplebed: I believe select statements choose which channel from which to read randomly when both have something available.  I expect them to block when neither does, and when either channel gets a message to accept it immediately. If all that's right, then the order of the two case statements shouldn't matter.

tail/tail.go
maplebed: because ... 
changed to return, removed the label. :)
(guessing it was a stuck mental pattern from the previous block)

leash.go
```diff
@@ -67,15 +74,36 @@ func run(options GlobalOptions) {
...
+		go func() {
+			select {
+			case <-sigs:
+				fmt.Fprintf(os.Stderr, "Caught second signal... Aborting.\n")
+				os.Exit(1)
+			case <-time.After(10 * time.Second):
+				fmt.Fprintf(os.Stderr, "Taking too long... Aborting.\n")
+				os.Exit(1)
+			}
+		}()
```
christineyen: aha!

ok. while calling a go func inside another go func isn't bad in and of itself, it **isn't** in fact necessary in this case!

your goal here was presumably to:
1. not block successful completion of this program by potentially waiting on a second signal or timeout
2. allow an exit valve if the "successful" completion takes too long after the first signal capture.

but! the fact that this is happening inside a goroutine in the first place means that #2 wouldn't happen, and since successful completion of this program means that everything in `linesChan` is closed and processed, the `exit 0` path is hit and so #1 is also irrelevant (we aren't forced, for example, to wait for the 10s timeout.)

so you can go straight from:

```go
close(abort)
select {
  case <-sigs:
    // ...
  case <-time.After(10 * time.Second):
    // ...
}
```

... and check all your boxes!

leash.go
```diff
@@ -67,15 +74,36 @@ func run(options GlobalOptions) {
...
 
+	// set up our signal handler, now that we know how many files we're tailing,
+	// we can send the right number of abort signals.
+	go func() {
+		sig := <-sigs
+		fmt.Fprintf(os.Stderr, "Aborting! Caught signal \"%s\"\n", sig)
+		fmt.Fprintf(os.Stderr, "Cleaning up...\n")
+		close(abort)
+		// and if they insist, catch a second CTRL-C or timeout on 10sec
+		signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
```
christineyen: https://husobee.github.io/golang/ecs/2016/05/19/ecs-graceful-go-shutdown.html

this is a cute pattern illustrating ^, in which their variant uses a channel to return the desired exit code (shrug, to each their own)

leash.go
maplebed: This is exactly where I was trying to go.  Thank you!  I'm glad we got there in the end.

PR #45, opened 2017-03-28
Adding a key=value parser

maplebed:
Basically copied the json parser wholesale, but used a library that reads log lines that look like
    key1=val1 key2=val2
instead of json to decode the lines.
Though it could use polish, it does appear to work. I'd rather get it in and building then improve so folks can try it out without building it themselves.

### DIFF
--- a/parsers/keyval/keyval.go
+++ b/parsers/keyval/keyval.go
@@ -51,6 +51,7 @@ func (p *Parser) Init(options interface{}) error {
 	return nil
 }
 
+// TODO dedupe the LineParser interface with the json parser or remove entirely
 type LineParser interface {
 	ParseLine(line string) (map[string]interface{}, error)
 }
@@ -130,6 +131,8 @@ func (p *Parser) ProcessLines(lines <-chan string, send chan<- event.Event, pref
 // if unable to parse it will return the current time
 // it is highly recommended that you populate the Config with time format
 // sample from logrus: "time":"2014-03-10 19:57:38.562264131 -0400 EDT"
+// TODO remove fancy time parsing from the keyval parser, since timestamps
+// are likely to be more well structured and come from the prefix
 func (p *Parser) getTimestamp(m map[string]interface{}) time.Time {
 	var ts time.Time
 	if p.conf.TimeFieldName != "" {
--- a/parsers/keyval/keyval.go
+++ b/parsers/keyval/keyval.go
@@ -86,7 +86,7 @@ func (p *Parser) ProcessLines(lines <-chan string, send chan<- event.Event, pref
 	for line := range lines {
 		logrus.WithFields(logrus.Fields{
 			"line": line,
-		}).Debug("Attempting to process json log line")
+		}).Debug("Attempting to process keyval log line")
 
 		// take care of any headers on the line
 		var prefixFields map[string]string
@@ -101,7 +101,8 @@ func (p *Parser) ProcessLines(lines <-chan string, send chan<- event.Event, pref
 		if err != nil {
 			// skip lines that won't parse
 			logrus.WithFields(logrus.Fields{
-				"line": line,
+				"line":  line,
+				"error": err,
 			}).Debug("skipping line; failed to parse.")
 			continue
 		}
@@ -120,7 +121,7 @@ func (p *Parser) ProcessLines(lines <-chan string, send chan<- event.Event, pref
 		}
 		send <- e
 	}
-	logrus.Debug("lines channel is closed, ending json processor")
+	logrus.Debug("lines channel is closed, ending keyval processor")
 }
 
 // getTimestamp looks through the event map for something that looks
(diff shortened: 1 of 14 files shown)


parsers/keyval/keyval.go
```diff
@@ -0,0 +1,216 @@
...
+	}
+	err := logfmt.Unmarshal([]byte(line), logfmt.HandlerFunc(f))
+	return parsed, err
+}
+
+func (p *Parser) ProcessLines(lines <-chan string, send chan<- event.Event, prefixRegex *parsers.ExtRegexp) {
+	for line := range lines {
+		logrus.WithFields(logrus.Fields{
+			"line": line,
+		}).Debug("Attempting to process json log line")
```
emfree: nit: s/json/keyvalue/

parsers/keyval/keyval.go
```diff
@@ -0,0 +1,216 @@
...
+		timestamp := p.getTimestamp(parsedLine)
+
+		// send an event to Transmission
+		e := event.Event{
+			Timestamp: timestamp,
+			Data:      parsedLine,
+		}
+		send <- e
+	}
+	logrus.Debug("lines channel is closed, ending json processor")
```
emfree: nit: s/json/keyvalue/

parsers/keyval/keyval.go
```diff
@@ -0,0 +1,216 @@
...
+
+func (p *Parser) Init(options interface{}) error {
+	p.conf = *options.(*Options)
+
+	p.nower = &RealNower{}
+	p.lineParser = &KeyValLineParser{}
+	return nil
+}
+
+type LineParser interface {
```
emfree: nit: maybe can put the LineParser interface in parsers/parsers.go instead of redefining in each of the subpackages?

parsers/keyval/keyval.go
```diff
@@ -0,0 +1,216 @@
...
+			line = strings.TrimPrefix(line, prefix)
+		}
+
+		parsedLine, err := p.lineParser.ParseLine(line)
+
+		if err != nil {
+			// skip lines that won't parse
+			logrus.WithFields(logrus.Fields{
+				"line": line,
+			}).Debug("skipping line; failed to parse.")
```
emfree: Maybe include the error value in the log statement as well?

parsers/keyval/timeformat.go
```diff
@@ -0,0 +1,66 @@
+package keyval
```
emfree: Would be nice to not duplicate this between `parsers/htjson` and `parsers/keyval`, but that can happen later :)

parsers/keyval/keyval.go
```diff
@@ -0,0 +1,216 @@
...
+// if unable to parse it will return the current time
+// it is highly recommended that you populate the Config with time format
+// sample from logrus: "time":"2014-03-10 19:57:38.562264131 -0400 EDT"
+func (p *Parser) getTimestamp(m map[string]interface{}) time.Time {
+	var ts time.Time
+	if p.conf.TimeFieldName != "" {
+		// remove the timestamp from the body when we stuff it in the header
+		defer delete(m, p.conf.TimeFieldName)
+		if t, found := m[p.conf.TimeFieldName]; found {
+			ts = p.tryTimeFormats(t.(string))
```
emfree: Is it definitely safe to cast t to a string at this point? What if the logfile has something like `timestamp=1490664217`?

parsers/keyval/keyval.go
maplebed: ehh, I'd rather not. The Parser interface is the slimmest connection - having a LineParser interface within the parser is actually not required at all.  Honestly I'm not sure why it's even here; there's only ever one implementation of the interface. (the MySQL parser, for example, doesn't have a LineParser anywhere.)

parsers/keyval/keyval.go
maplebed: done

parsers/keyval/keyval.go
maplebed: done

parsers/keyval/keyval.go
maplebed: done

parsers/keyval/timeformat.go
maplebed: yeah...  and I think more likely is to strip it out of keyval entirely - whereas with a JSON line the timefield is almost guaranteed to be buried in a field somewhere, I'm guessing that a keyval line is more likely to have it in the prefix, and we can easily make stronger requirements about the field name (if not format) for that.

parsers/keyval/keyval.go
maplebed: whoop!  
➜  go test ./...
```
--- FAIL: TestGetTimestampValid (0.00s)
panic: interface conversion: interface {} is int, not string [recovered]
        panic: interface conversion: interface {} is int, not string
```

maplebed: fixed

parsers/keyval/keyval.go
christineyen: +1 eventually moving away from the (exported!) duplicate `LineParser` interfaces everywhere. doesn't have to be in this PR, but it doesn't seem to have much of a purpose anymore and seems to be a victim/side effect of frequent copypasta.

PR #46, opened 2017-03-29
Add flag to keyval parser to filter incoming mesasge stream

maplebed:
All other parsers have strict conditions under which they'll try and parse a line.  The keyval parser doesn't, and so extra lines in the log file (that don't have a key=val format) cause extra fields to be created, to disastrous effect. 
This change adds a flag to let you specify a regular expression to require that only matching lines get parsed or, when the inverse flag is added, to parse all lines except those that match.

### DIFF
--- a/parsers/keyval/keyval.go
+++ b/parsers/keyval/keyval.go
@@ -121,13 +121,20 @@ func (p *Parser) ProcessLines(lines <-chan string, send chan<- event.Event, pref
 		}
 
 		parsedLine, err := p.lineParser.ParseLine(line)
-
 		if err != nil {
 			// skip lines that won't parse
 			logrus.WithFields(logrus.Fields{
 				"line":  line,
 				"error": err,
 			}).Debug("skipping line; failed to parse.")
+			continue
+		}
+		if len(parsedLine) == 0 {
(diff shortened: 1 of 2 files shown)


christineyen:
i would also like some sort of failsafe where... if *no* key/value pairs are found, the entire row is thrown away. something that prevents this sort of garbage:

```
116.251.218.99:40810 [28/Mar/2017:20:14:41.839] live_http_protocol/6: Connection closed during SSL handshake
```

from **ever** being considered a valid k/v line. however it makes sense for that to happen.

parsers/keyval/keyval.go
```diff
@@ -137,6 +137,15 @@ func (p *Parser) ProcessLines(lines <-chan string, send chan<- event.Event, pref
 			}).Debug("skipping line; no key/val pairs found.")
 			continue
 		}
+		if allEmpty(parsedLine) {
+			// skip events for which all fields are the empty string, because that's
+			// probably broken
+			logrus.WithFields(logrus.Fields{
+				"line":  line,
+				"error": err,
+			}).Debug("skipping line; no all values are the empty string.")
```
emfree: nit: s/no all/all/

PR #47, opened 2017-03-30
thicken pipe between the the parser and libhoney, enable multilpe parallel parsers

maplebed:
Once the tailer reads a line, it hands it off to the parser. The parser sends it through a whole chain of channels to get adjusted before it gets to libhoney.
The disk is not the bottleneck, and libhoney is getting batching, removing it as the bottleneck. That leaves the parser and all the munging parts.
This change configures all the munging channels to use as many goroutines as the senders, so that CPU intensive things like calculating SHA hashes or parsing URLs into constituent parts can be done in parallel, if you have the CPUs for it.
Additionally, it sets each of the parsers to run 20 goroutines, in hopes that they will be able to more efficiently drain the channel coming from the tailer. The parsers all treat each line as an independent event, so there is no danger (except maybe mixing up the final order a bit, which is fine) in increasing the number of workers doing the heavy lifting of the parser.

### DIFF
--- a/leash.go
+++ b/leash.go
@@ -210,47 +210,39 @@ func getParserAndOptions(options GlobalOptions) (parsers.Parser, interface{}) {
 // returns a channel on which it will send the munged events.
 // It is responsible for hashing or dropping or adding fields to the events
 func modifyEventContents(toBeSent chan event.Event, options GlobalOptions) chan event.Event {
-	for _, field := range options.DropFields {
-		toBeSent = dropEventField(field, toBeSent, options)
+	// short circuit this if no field scrubbing is enabled
+	if len(options.DropFields) == 0 && len(options.ScrubFields) == 0 &&
+		len(options.AddFields) == 0 && len(options.RequestShape) == 0 {
+		return toBeSent
 	}
-	for _, field := range options.ScrubFields {
-		toBeSent = scrubEventField(field, toBeSent, options)
-	}
-	for _, field := range options.AddFields {
-		toBeSent = addEventField(field, toBeSent, options)
-	}
-	for _, field := range options.RequestShape {
-		toBeSent = requestShape(field, toBeSent, options)
-	}
-	return toBeSent
-}
-
-// dropEventField drops any fields that are to be dropped, drop them before
-// passing the event on down the line to the next consumer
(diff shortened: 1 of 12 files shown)


maplebed:
whoops, broke tests that probably depend on the order of things that are now unordered. :/

christineyen:
seems like the tests should stop depending on ordering! :D

leash.go
```diff
@@ -218,13 +226,21 @@ func modifyEventContents(toBeSent chan event.Event, options GlobalOptions) chan
 
 // dropEventField drops any fields that are to be dropped, drop them before
 // passing the event on down the line to the next consumer
-func dropEventField(field string, toBeSent chan event.Event) chan event.Event {
-	newSent := make(chan event.Event)
+func dropEventField(field string, toBeSent chan event.Event, options GlobalOptions) chan event.Event {
```
christineyen: Hokay. Remind me why this needed to be written as a transformation on a channel? There's no reason that `dropEventField` / `scrubEventField` / `addEventField` / `requestShape` need to be serialized like that. They **do** probably need to operate on the same event, within a single goroutine... but that seems simple enough to do.

I don't think making this more complicated is the answer.

leash.go
```diff
@@ -133,9 +133,17 @@ func run(options GlobalOptions) {
...
+			for i := uint(0); i < options.NumSenders; i++ {
+				wg.Add(1)
+				go func() {
+					for ev := range modifiedToBeSent {
+						realToBeSent <- ev
+					}
+					wg.Done()
+				}()
 			}
+			wg.Wait()
```
christineyen: See comment below in `dropEventField`, first.

I don't think this is what needs to happen, and I don't think we need `toBeSent`, `modifiedToBeSent`, *and* `realToBeSent`. I think we need one of them.

Pass `options` down into `sendToLibhoney`, and simplify `modifyEventContents` to take an `event.Event` instead of a `chan event.Event` and just... modify it, no return value. no channels, no waitgroups. Everything you're doing to the event is simple transformation, there's no reason for the *up to N* channels involved in things. And `modifyEventContents` can just be called from inside the `sendToLibhoney` loop when pulling off `toBeSent`.

**Simplify**. `honeytail` does lots of things but in the end it's a log parser/processor/transmitter, and most of it should be able to be expressed fairly straightforwardly.

PR #49, opened 2017-05-03
Add journalctl-based parser

sandalwing:
cc @christineyen @samstokes @maplebed @toshok @charity 

Send your journals to Honeycomb. Systemd is the best init in the world no imitators


sandalwing:
Hold off on merging please because I want to rename to `journald` to be proper maybe? But PTAL

Issue #53, opened 2017-06-21
honeytail doesn't like nested JSON


nathanleclaire:
Even though we support it `honeytail` will simply string encode nested JSON in the parser today!

see for reference: https://github.com/honeycombio/honeytail/blob/master/parsers/htjson/htjson.go#L65-L87

nathanleclaire:
Fixed!

PR #54, opened 2017-06-21
update json parser to pass through nested objects

maplebed:
Honeycomb used to drop incoming events that had nested JSON. It no longer does, so this change passes through JSON objects that have any kind of nesting instead of stringifying the nested structure.
Fixes issue #53


nathanleclaire:
+7, -21 lines of code -- <3

LGTM

PR #59, opened 2017-07-14
Change default below which dynsampling is disabled, allow override

maplebed:
The dynamic sampling algorithm used in honeytail is [Average with Minimum](https://godoc.org/github.com/honeycombio/dynsampler-go#AvgSampleWithMin). The default minimum throughput under which sampling is disabled is 50. This is too high a default for honeytail, where (in a scaled out system) each host may be handling only tens of requests per second. 

This change adjusts the default to 1 because it's kind of weird to enable dynsampling but not see it take effect. For the folks that dig in to their system and really want to have a high minimum threshold, it adds a flag to set the minimum.

### DIFF
--- a/leash.go
+++ b/leash.go
@@ -263,6 +263,7 @@ func modifyEventContents(toBeSent chan event.Event, options GlobalOptions) chan 
 		sampler = &dynsampler.AvgSampleWithMin{
 			GoalSampleRate:    options.GoalSampleRate,
 			ClearFrequencySec: options.DynWindowSec,
+			MinEventsPerSec:   options.MinSampleRate,
 		}
 		if err := sampler.Start(); err != nil {
 			logrus.WithField("error", err).Fatal("dynsampler failed to start")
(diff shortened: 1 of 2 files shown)


samstokes:
This shouldn't block shipping the change if you feel strongly about getting it out, but I'm not sure I understand the problem this is trying to solve.

I'm not very familiar with the problem that the minimum throughput threshold is trying to solve ("the failings it shows on the low end of total traffic throughput"). Is it a question of risking throwing away too much signal during periods of low throughput?

If that is the idea, then it seems like this change might introduce a footgun for high-volume customers (i.e. changes the default behaviour to something they probably won't notice until it bites them).

What is the current flow by which customers "see sampling take effect" after enabling it? I didn't think we exposed sample rate as a field you could graph in the UI. Are they looking at the throughput graph in the usage panel and wondering why it didn't go down?

If the right behaviour fundamentally depends on the nature of the service being monitored (e.g. scaled-out services are more likely to want a low minimum, log aggregators and Big Java Monoliths want a high minimum), then config makes sense; but maybe the best way to get people to choose the right config for their situation is with docs. They're pretty unlikely to enable dynamic sampling without consulting the docs anyway, so a CAVEATS section stands a reasonable chance of being read.

maplebed:
Your guess about the reason for this is right on. If your traffic has a very high fluctuation between peak and trough, you will throw away too much signal during the trough.

None of our current honeytail users depend on this (or, I expect, even know it happens). The only people this change will affect are those that want sampling below the current minimum. 

You're right, we don't make it easy for folks to see the sample rate affecting their traffic. This change doesn't affect that, though it might encourage us to pass on that info in a nicer way.

I'll land this, because I think all your questions are better addressed outside this repo (docs, UI tweaks, etc.).

PR #62, opened 2017-08-16
Add function to convert to HT's flag.Filename

nathanleclaire:
cc @maplebed @emfree 

It might seem weird, but it's needed to use `nginx.Parser` in https://github.com/honeycombio/honeyelb. If you try to do a type conversion to `flag.Filename` in `honeyelb` and pass it to `honeytail` it won't compile because these two modules are different:

- `github.com/honeycombio/honeytail/vendor/github.com/jessevdk/go-flags`
- `github.com/honeycombio/honeyelb/vendor/github.com/jessevdk/go-flags`

 are different import paths (and therefore different types as far as Go is concerned). And you can't import `vendor/` packages from another code repo directly, it's not allowed.

Open to ideas but this seems the fastest route.

Signed-off-by: Nathan LeClaire <nathan@honeycomb.io>

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -1,0 +1,7 @@
+package types
+
+import flags "github.com/jessevdk/go-flags"
+
+func FilenameFlagFromString(str string) flags.Filename {
+	return flags.Filename(str)
+}


emfree:
Do you have a patch in honeyelb that shows the compiler error? I think I sort of understand what's going on, but it'd be helpful to see the whole picture.

emfree:
I see. :thinking: What if we made `nginx.Options.ConfigFile` simply have type `string`? Alternatively: if we vendor honeytail in honeyelb and flatten nested `vendor` directories, the type incompatibility should go away.

nathanleclaire:
> What if we made nginx.Options.ConfigFile simply have type string

That'd be OK with me, just not sure what special properties the `flags.Filename` type has as far as flag parsing goes.

> Alternatively: if we vendor honeytail in honeyelb and flatten nested vendor directories, the type incompatibility should go away

Is that functionality built in to `glide`? I'm not well versed on the finer details of how nested `vendor/` dirs work

nathanleclaire:
looks like flattening `vendor` seems possible https://stackoverflow.com/questions/38597046/how-to-handle-nested-vendor-directories-in-go-packages

nathanleclaire:
I do intend to vendor `github.com/honeycombio/honeytail` in `honeyelb` anyway.

nathanleclaire:
Completion (i.e. identifying which files match a `*pattern*`) does seem one of the main goals of `flag.Filename` (from `github.com/jessevdk/go-flags/completion.go`):

```
61 // Complete returns a list of existing files with the given                                                                                                                                                                                                                                                                                      
 62 // prefix.                                                                                                                                                                                                                                                                                                                                       
 63 func (f *Filename) Complete(match string) []Completion {                                                                                                                                                                                                                                                                                         
 64         ret, _ := filepath.Glob(match + "*")                                                                                                                                                                                                                                                                                                     
 65         return completionsWithoutDescriptions(ret)                                                                                                                                                                                                                                                                                               
 66 }  
```

I don't think the nginx parser would want more than one file, so maybe changing to `string` is simplest in interim?

nathanleclaire:
closing for #63

PR #65, opened 2017-08-31
Don't use regex to match MySQL queries

emfree:
Previously, when parsing a MySQL query string, the MySQL parser would match a
line up to the first '#' character with the `reQuery` regular expression.

The intent must be to strip comments; however, this strategy fails when parsing
valid queries such as
```
SELECT * FROM images WHERE hashtag="#ballin";
```

or when parsing query strings that include a '#' character in a C-style comment
block:
```
SELECT 1 /* callsite:class#method */;
```
etc.

Instead, treat the whole line as part of the query string. It is true that this
doesn't properly handle query strings of the form
```
SELECT * FROM users # comment here
    WHERE id=1;
```
However, only query strings from tcpdump or similar retain comments (the actual
slow query log strips them), so that seems like a reasonable omission.

### DIFF
--- a/parsers/mysql/mysql_test.go
+++ b/parsers/mysql/mysql_test.go
@@ -483,7 +483,7 @@ func init() {
 			},
 			timestamp: time.Unix(1476901800, 0),
 		},
-		{ /* 25 */
+		{ /* 24 */
 			rawE: []string{
 				"# User@Host: rdsadmin[rdsadmin] @ localhost [127.0.0.1]  Id:     1",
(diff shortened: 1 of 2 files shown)


parsers/mysql/mysql_test.go
```diff
@@ -483,6 +483,47 @@ func init() {
 			},
 			timestamp: time.Unix(1476901800, 0),
 		},
+		{ /* 25 */
```
maplebed: nit - s/25...24/24...25/ in counting test cases

PR #66, opened 2017-08-31
accept old style timestamp and truncated query duration line

maplebed:


parsers/mysql/mysql.go
```diff
@@ -99,9 +99,11 @@ const (
 
 var (
 	reTime             = parsers.ExtRegexp{regexp.MustCompile("^# Time: (?P<time>[^ ]+)Z *$")}
+	reVCTime           = parsers.ExtRegexp{regexp.MustCompile("^# Time: (?P<datetime>[0-9]+ [0-9:]+)")}
```
emfree: nit: maybe add a comment explaining why we have these alternate regexes? (VC=vividcortex?)

maplebed: ::sigh:: more accurate would be pre-mysql-5.somethingerother... I'll change the name to reflect that.

PR #69, opened 2017-09-28
add a --localtime and --timezone flag to control time parsing

maplebed:
Previously honeytail would always parse timestamps with no time zone as though they were UTC. This change adds two flags to change that behavior:
* --localtime will parse untimezoned timestamps as though they were in the same time zone as the machine doing the parsing
* --timezone takes a time zone argument (eg `America/New_York`) and parses untimezoned timestamps as though they were in that time zone.


httime/httime.go
```diff
@@ -172,25 +181,25 @@ func tryTimeFormats(t, intendedFormat string) time.Time {
...
 		// Still try Go style, just in case
-		if ts, err := time.Parse(format, t); err == nil {
+		if ts, err := time.ParseInLocation(format, t, Location); err == nil {
 			return ts
 		}
 	}
 
 	var ts time.Time
-	if tOther, err := time.Parse("2006-01-02 15:04:05.999999999 -0700 MST", t); err == nil {
+	if tOther, err := time.ParseInLocation("2006-01-02 15:04:05.999999999 -0700 MST", t, Location); err == nil {
```
emfree: could instead call `tOther, err := Parse(format, timespec); ...` here and below

PR #58, opened 2017-06-29
Use common LineParser interface in individual parsers.

emfree:
Previously, each parser package defined a more-or-less duplicated `LineParser`
interface. Unify them to all implement the same interface.

This doesn't change honeytail behavior, but makes it easier to reuse individual
parsers in the fledgling kubernetes agent.

Test plan: Unit tests

PR #72, opened 2017-11-07
 Add Postgres slow query parser

emfree:
This is a fairly minimal stab at slurping Postgres query logs. In Postgres,
slow queries go into the general log if you set the
`log_min_duration_statement` setting, and look like this:
```
2017-11-07 01:43:39 UTC [3542-7] postgres@test LOG:  duration: 15.577 ms  statement: SELECT * FROM test WHERE id=1;
```

There are a number of things that I've omitted from this patch for the moment,
in the interest of simplicity:

- Running multiple parser goroutines in parallel.
- Propagating a sample rate down to the parser.
- Verifying compatibility across postgres versions -- I've only tested this
  with PostgreSQL 9.5.
- Parsing log statements that are not slow queries.

I plan to address the first three in subsequent patches, but might punt on
parsing other types of log statements unless there's demonstrated interest.

Note: this looks like a monster diff, but it includes (and depends on) #58 and #71. Only the latest commit is the relevant one.

### DIFF
--- a/parsers/postgresql/postgresql_test.go
+++ b/parsers/postgresql/postgresql_test.go
@@ -12,11 +12,13 @@ import (
 // Test parsing individual log statements with different prefix formats.
 func TestSingleQueryParsing(t *testing.T) {
 	testcases := []struct {
+		description  string
 		in           string
 		prefixFormat string
 		expected     event.Event
 	}{
 		{
+			description: "parse a multiline log statement from a default postgres 9.5 log",
 			in: `2017-11-07 00:05:16 UTC [3053-3] postgres@postgres LOG:  duration: 0.681 ms  statement: SELECT d.datname as "Name",
 	       pg_catalog.pg_get_userbyid(d.datdba) as "Owner",
 	       pg_catalog.pg_encoding_to_char(d.encoding) as "Encoding",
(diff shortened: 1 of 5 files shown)


emfree:
An update, before you start digging into this: In this patch, I'm parsing the line prefix (`2017-11-07 01:43:39 UTC [3542-7] postgres@test`) using a hard-coded regular expression. Predictably, this turns out to be a delusionally optimistic strategy. The line prefix:
- varies significantly between minor versions
- is customizable by the user using an esoteric and underspecified [format](https://github.com/postgres/postgres/blob/f49842d1ee31b976c681322f76025d7732e860f3/src/backend/utils/misc/postgresql.conf.sample#L445-L465)

So this will need a bit more work. Stay tuned....

parsers/postgresql/postgresql_test.go
```diff
@@ -0,0 +1,150 @@
...
+	"testing"
+	"time"
+
+	"github.com/honeycombio/honeytail/event"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestSingleQueryParsing(t *testing.T) {
+	testcases := []struct {
+		in           string
```
maplebed: could you add a `description` field here that describes what the test is looking for? then in the testing below, include that field in the output when a test fails.

PR #73, opened 2017-11-09
Regex log line parser

rfong:
Log parser that uses arbitrary user-defined regexes

Example usage from command line:
```
go install . && \
	honeytail -p regex -k $HONEYTAIL_WRITEKEY \
	-f /private/var/system.log \
	--dataset 'MY_TEST_DATASET' \
	--backfill \
	--regex.line_regex="(?P<time>\d{2}:\d{2}:\d{2})" \
	--regex.timefield="time" \
	--regex.time_format="%H:%M:%S"
```

More elaborate parsing examples in the unit tests ^_^


parsers/regex/regex.go
```diff
@@ -0,0 +1,158 @@
...
+
+type Parser struct {
+	conf       Options
+	lineParser LineParser
+}
+
+func (p *Parser) Init(options interface{}) error {
+	p.conf = *options.(*Options)
+
+	if p.conf.LineRegex != "" {
```
emfree: Do we also want to consider it an error if `LineRegex` is empty here? Seems like it wouldn't be very useful to run honeytail with `--parser=regex` if you don't also specify a regex.

parsers/regex/regex.go
```diff
@@ -0,0 +1,158 @@
...
+			}).Debug("Could not compile line regex")
+			return err
+		}
+		p.lineParser = &RegexLineParser{lineRegex}
+	}
+
+	return nil
+}
+
+type LineParser interface {
```
emfree: Nit: we just recently merged #58, which added this interface definition in the top-level `parsers` package. If you rebase onto master, you can reference `parsers.LineParser` instead of needing to redefine it here.

parsers/regex/regex.go
```diff
@@ -0,0 +1,158 @@
...
+	if match == nil || len(match) == 0 {
+		logrus.WithFields(logrus.Fields{
+			"line": line,
+		}).Debug("No matches for regex log line")
+		return parsed, nil
+	}
+
+	// Map capture groups
+	var firstMatch []string = match[0] // We only care about the first full lineRegex match
+	for i, name := range p.lineRegex.SubexpNames() {
```
emfree: If you want, I think you can skip some of this rigamarole by using the `ExtRegexp` type from the `parsers` package in honeytail. It just wraps a `*regexp.Regexp` and adds a `FindStringSubmatchMap` method, which I believe is pretty similar to what's happening here. (Though functionally, this code looks good, so up to you ¯\_(ツ)_/¯)

parsers/regex/regex.go
```diff
@@ -0,0 +1,158 @@
...
+				send <- e
+			}
+			wg.Done()
+		}()
+	}
+	wg.Wait()
+	logrus.Debug("lines channel is closed, ending regex processor")
+}
+
+type NoopLineParser struct {
```
emfree: Unused?

parsers/regex/regex_test.go
```diff
@@ -0,0 +1,168 @@
...
+		expected: map[string]interface{}{
+			"time":    "2017/11/07 22:59:46",
+			"status":  "error",
+			"request": "GET /isbns HTTP/1.1",
+		},
+	},
+}
+
+func TestParseLine(t *testing.T) {
+	p := &Parser{
```
emfree: It might be nice to call `p.Init(Options{...})` instead of populating the struct members directly, in order to also test that `Init()` does the right thing. (And maybe also worth adding a test to check that `Init()` fails sensibly if you give it a bad regex)

parsers/regex/regex.go
```diff
@@ -0,0 +1,158 @@
...
+
+	"github.com/Sirupsen/logrus"
+
+	"github.com/honeycombio/honeytail/event"
+	"github.com/honeycombio/honeytail/httime"
+	"github.com/honeycombio/honeytail/parsers"
+)
+
+type Options struct {
+	LineRegex       string `long:"line_regex" description:"a regular expression with named capture groups representing the fields you want parsed"`
```
emfree: Let's make sure to add some documentation and examples to the website, and link to it from this help string. Capture group syntax varies between languages, so this could be confusing otherwise.

parsers/regex/regex_test.go
```diff
@@ -0,0 +1,168 @@
...
+
+type testLineMaps struct {
+	line        string
+	trimmedLine string
+	resp        map[string]interface{}
+	typedResp   map[string]interface{}
+	ev          event.Event
+}
+
+type FakeLineParser struct {
```
emfree: Unused?

parsers/regex/regex.go
```diff
@@ -0,0 +1,158 @@
...
+type Parser struct {
+	conf       Options
+	lineParser LineParser
+}
+
+func (p *Parser) Init(options interface{}) error {
+	p.conf = *options.(*Options)
+
+	if p.conf.LineRegex != "" {
+		var err error
```
maplebed: this declaration is unnecessary - the following line can declare both `lineRegex` and `err`

parsers/regex/regex.go
```diff
@@ -0,0 +1,158 @@
...
+		lineRegex, err := regexp.Compile(p.conf.LineRegex)
+		if err != nil {
+			logrus.WithFields(logrus.Fields{
+				"LineRegex": p.conf.LineRegex,
+			}).Debug("Could not compile line regex")
+			return err
+		}
+		p.lineParser = &RegexLineParser{lineRegex}
+	}
```
maplebed: minor hand-wringing, but I wonder if we should do some more heuristic testing of the regex they hand us to ensure things like at least one named group exists. Feels terrible to do string parsing on the regex itself to test for these sorts of things...

parsers/regex/regex.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+type RegexLineParser struct {
+	lineRegex *regexp.Regexp
+}
+
+func (p *RegexLineParser) ParseLine(line string) (map[string]interface{}, error) {
+	parsed := make(map[string]interface{})
+	if p.lineRegex == nil {
+		logrus.Debug("RegexLineParser", p, "has no lineRegex")
```
maplebed: This feels more like `logrus.Error` than `Debug`. I suppose it should only happen in tests though, right? since honeytail runs init, which ensures it's not nil? Yeah, for sure Error then because it can only get hit when someone's using the parser in some other way.

parsers/regex/regex_test.go
```diff
@@ -0,0 +1,168 @@
...
+		lineRegex: `(?P<outer>[^ ]* (?P<inner1>[^ ]*) (?P<inner2>[^ ]*))`,
+		input:     `foo bar baz`,
+		expected: map[string]interface{}{
+			"outer":  "foo bar baz",
+			"inner1": "bar",
+			"inner2": "baz",
+		},
+	},
+	{
+		// Sample nginx log line
```
maplebed: nit: please clarify to `nginx error log line` to differentiate it from the `access.log` line that the nginx parser usually gets.

parsers/regex/regex_test.go
```diff
@@ -0,0 +1,168 @@
...
+			ev: event.Event{
+				Timestamp: t1,
+				Data: map[string]interface{}{
+					"http_x_forwarded_proto": "https",
+					"remote_addr":            "10.252.4.24",
+				},
+			},
+		},
+	}
+	lineRegex, err := regexp.Compile(`(?P<http_x_forwarded_proto>\w+) - (?P<remote_addr>\d{1,4}\.\d{1,4}\.\d{1,4}\.\d{1,4}) - - \[(?P<local_time>\d{2}\/[A-Za-z]+\/\d{4}:\d{2}:\d{2}:\d{2}.*)\]`)
```
maplebed: I'm confused - how is this test passing? The named group is `local_time` but the `resp` map above lists `time_local` as the expected key.
oh! `pair.resp` isn't actually used anywhere - the `resp` down below is the parsed line, not the map from the test definition. 
Should we drop `resp` from the test setup or add a check down below to check it? I lean towards dropping it from the test definition since this test is using the `ProcessLines` interface rather than the `LineParser` so won't have access to the map, only the event.

parsers/regex/regex_test.go
rfong: weird, none of the other parser test files appear to test for `p.Init(...)` success

I'll fix it in this file and add a note to look thru the rest of honeytail in a separate diff :)

parsers/regex/regex.go
rfong: whoops typo on my part! good catch

parsers/regex/regex.go
rfong: no need, golang `regexp` has a thing for that ^_^

parsers/regex/regex.go
```diff
@@ -0,0 +1,165 @@
...
+	lineParser parsers.LineParser
+}
+
+func (p *Parser) Init(options interface{}) error {
+	p.conf = *options.(*Options)
+
+	// Regex can't be blank
+	if p.conf.LineRegex == "" {
+		logrus.Debug("LineRegex is blank")
+		return errors.New("LineRegex is blank")
```
emfree: Small request: this error gets bubbled up to users if they don't specify the `--regex.line_regex` flag. Can we make it a little clearer? Right now, running honeytail without specifying a regex prints:
```
$ honeytail -f - -k $WRITEKEY -d test -p regex
INFO[0000] Starting honeytail                           
FATA[0000] err initializing parser module                err=LineRegex is blank parser=regex
```
which would probably be a bit opaque to the typical user.

parsers/regex/regex.go
```diff
@@ -0,0 +1,165 @@
...
+		for _, groupName := range lineRegex.SubexpNames() {
+			if groupName != "" {
+				numNamedGroups++
+			}
+		}
+		if numNamedGroups == 0 {
+			logrus.WithFields(logrus.Fields{
+				"LineRegex": p.conf.LineRegex,
+			}).Error("No named capture groups")
+			return errors.New("No named capture groups found; must provide at least one named group")
```
emfree: Here again, I wonder if we can make this clearer for users. Right now this'll print 
```
INFO[0000] Starting honeytail                           
ERRO[0000] No named capture groups                       LineRegex=.*
FATA[0000] err initializing parser module                err=No named capture groups found; must provide at least one named group parser=regex
```

which is a bit confusing to me. Could we return an error value `No named capture groups found in the regex <regex>. Your regex must contain at least one named group`? Or something like that?

parsers/regex/regex_test.go
```diff
@@ -0,0 +1,222 @@
+package regex
+
+import (
+	//"errors"
```
emfree: nit: stray commented code

PR #74, opened 2017-11-09
Use `context` instead of our own `abort` channel to signal

christineyen:
Precursor work to some honeytail telemetry work -- use `context.Context`s instead.

I also added a `<-ctx.Done():` inside `sendToLibhoney` so that we're not hanging on clearing a send queue when we get a signal. (Happy to be pushed back with the concern that we'd drop events on the floor in the case of an intentional pause / intent to resume, but it feels like most folks at this point are just `CTRL-C`'ing and want the thing to stop. Shrug.)

### DIFF
--- a/leash.go
+++ b/leash.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"crypto/sha256"
 	"fmt"
 	"math/rand"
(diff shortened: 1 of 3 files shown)


leash.go
```diff
@@ -434,6 +434,9 @@ func sendToLibhoney(toBeSent chan event.Event, toBeResent chan event.Event,
 		}
 		// otherwise pick something up off the regular queue and send it
 		select {
+		case <-ctx.Done():
```
maplebed: I think I'd rather leave this one out. When the CTRL-C comes in, it will close the tailer which will in turn close lines and we'll hit the closed channel once all the in-flight events are sent. I think this fits better with the expectation set about picking up tailing from where we left off. There is also the 10sec timer to really kill it if the queue is full enough that it can't get them out in that time, so adding this only changes time to exit my a max of 10sec.

christineyen: Nod. I was iffy about the guarantees around resuming, but you're right and impatient debug cyen of the past should just shove it :)

PR #75, opened 2017-11-09
Support opt-in telemetry sent to a `-reports` dataset

christineyen:
Influenced by #28, but is opt-in and richer than the summary stats proposed in that diff.

The goal is to allow folks to capture richer telemetry from honeytail on an opt-in basis, to a dataset within the same team. (Suffixed with `-reports`, and relies on the currently-default automatic-creation behavior.)

The existence of a `libhoney.Builder` (which in turns is determined by the presence of the `--report` flag) dictates whether a telemetry event gets sent, and we use dynamic sampling to keep those telemetry events under control.

This diff also does a pass to clean up and standardize existing `.Debug` lines that line up with `reporting.*` calls.

### DIFF
--- a/reporting/reporting.go
+++ b/reporting/reporting.go
@@ -2,6 +2,7 @@ package reporting
 
 import (
 	"context"
+	"os"
 	"strings"
 	"time"
 
@@ -21,9 +22,13 @@ const (
 func NewContext(ctx context.Context) context.Context {
 	builder := libhoney.NewBuilder()
 	builder.Dataset += reportingSuffix
+	if hostname, err := os.Hostname(); err == nil {
+		builder.AddField("hostname", hostname)
+	}
+
 	sampler := &dynsampler.PerKeyThroughput{
 		ClearFrequencySec:      1,
-		PerKeyThroughputPerSec: 1000,
+		PerKeyThroughputPerSec: 10,
 	}
 	ctx = context.WithValue(ctx, contextKeyBuilder, builder)
 	if err := sampler.Start(); err != nil {
--- a/reporting/reporting.go
+++ b/reporting/reporting.go
@@ -2,7 +2,6 @@ package reporting
 
 import (
 	"context"
-	"encoding/json"
 	"strings"
 	"time"
 
@@ -19,16 +18,13 @@ const (
 
 func NewContext(ctx context.Context) context.Context {
 	builder := libhoney.NewBuilder()
+	builder.Dataset += reportingSuffix
 	return context.WithValue(ctx, contextKeyBuilder, builder)
 }
 
 func Options(ctx context.Context, options interface{}) {
 	if ev := getEvent(ctx); ev != nil {
-		optj, err := json.Marshal(options)
-		if err != nil {
-			return
-		}
-		ev.AddField("config_json", string(optj))
+		ev.AddField("config_json", options)
 		ev.Send()
 	}
 }
@@ -121,7 +117,5 @@ func getEvent(ctx context.Context) *libhoney.Event {
 		return nil
 	}
 
-	ev := builder.NewEvent()
-	ev.Dataset += reportingSuffix
-	return ev
+	return builder.NewEvent()
 }
--- a/reporting/reporting.go
+++ b/reporting/reporting.go
@@ -6,31 +6,42 @@ import (
 	"time"
 
 	"github.com/Sirupsen/logrus"
+	"github.com/honeycombio/dynsampler-go"
 	"github.com/honeycombio/libhoney-go"
 
 	"github.com/honeycombio/honeytail/event"
 )
 
 const (
-	contextKeyBuilder = "reporting"
+	contextKeyBuilder = "builder"
+	contextKeySampler = "sampler"
 	reportingSuffix   = "-reports"
 )
 
 func NewContext(ctx context.Context) context.Context {
 	builder := libhoney.NewBuilder()
 	builder.Dataset += reportingSuffix
-	return context.WithValue(ctx, contextKeyBuilder, builder)
+	sampler := &dynsampler.PerKeyThroughput{
+		ClearFrequencySec:      1,
+		PerKeyThroughputPerSec: 1000,
+	}
+	ctx = context.WithValue(ctx, contextKeyBuilder, builder)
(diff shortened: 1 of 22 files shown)


reporting/reporting.go
```diff
@@ -0,0 +1,127 @@
...
+	return context.WithValue(ctx, contextKeyBuilder, builder)
+}
+
+func Options(ctx context.Context, options interface{}) {
+	if ev := getEvent(ctx); ev != nil {
+		optj, err := json.Marshal(options)
+		if err != nil {
+			return
+		}
+		ev.AddField("config_json", string(optj))
```
maplebed: why marshal and stringify options instead of adding them as an object and letting libhoney do it?

reporting/reporting.go
```diff
@@ -0,0 +1,127 @@
...
+	SkipWithFields(ctx, line, skipReason, nil)
+}
+
+func SkipWithFields(ctx context.Context, line, skipReason string, fields logrus.Fields) {
+	logrus.WithFields(fields).WithField("line", line).Debugln("Skipped:", skipReason)
+
+	if ev := getEvent(ctx); ev != nil {
+		ev.AddField("log_line", line)
+		ev.AddField("skip_reason", skipReason)
+		ev.Send()
```
maplebed: could we add `ev.Add(fields)` too?

reporting/reporting.go
```diff
@@ -0,0 +1,127 @@
...
+		sentEvent := rsp.Metadata.(event.Event)
+		ev.AddField("event_timestamp", sentEvent.Timestamp)
+		ev.AddField("event_timestamp_lag_sec", time.Since(sentEvent.Timestamp)/time.Second)
+		ev.AddField("event_data", sentEvent.Data)
+		ev.AddField("response_status_code", rsp.StatusCode)
+		ev.AddField("response_latency_ms", rsp.Duration/time.Millisecond)
+		if rsp.StatusCode > 400 {
+			ev.AddField("honeycomb_error", rsp.Err.Error())
+		}
+		ev.Send()
```
maplebed: of all the spots you're thinking of adding dynsampling, this feels like the spot it'll be needed most.

reporting/reporting.go
```diff
@@ -0,0 +1,127 @@
...
+	"github.com/honeycombio/honeytail/event"
+)
+
+const (
+	contextKeyBuilder = "reporting"
+	reportingSuffix   = "-reports"
+)
+
+func NewContext(ctx context.Context) context.Context {
+	builder := libhoney.NewBuilder()
```
maplebed: I feel like there should be some things shoved in to the builder here, but I'm not sure what. This tails into our discussion about this dataset as a grab bag of junk...
ooh! If you do tie in to the stats at all, a dynamic function representing the number of events processed since the last message would be kinda neat.

maplebed: Any particular reason you chose to put the dataset name munging in `getEvent` instead of here?

reporting/reporting.go
christineyen: Ahh, I forgot that [our `libhoney.Add`](https://github.com/honeycombio/libhoney-go/blob/master/libhoney.go#L408) respects JSON tags, I had the impression that it just used reflection.

I was mostly hoping to rely on `omitempty` to make the config file less noisy. 😶 

Ahhhhh I see, but we're not talking about `.Add()` at all, are we? And `AddField("config", theconfigobj)` does the right thing?

🙏  grateful, thanks

reporting/reporting.go
christineyen: Oops, nope, that was a bridge lost in the iteration swamp.

reporting/reporting.go
christineyen: I turned up my nose at the `fields` because they a) either repeat data that already exists (e.g. `line` and `rawE`, or `samplerate` in the config) or b) are... redundant (e.g. the value of `matched`, which can be otherwise determined by the value of `InvertFilter` and this log line being output.

But I see your hoping for consistency. Would you be horrified if I instead axed those additional fields from those log lines? 😶

reporting/reporting.go
christineyen: Yep. TBH some `PerKeyThroughput` should be able to be applied to all of them, expecting to be a noop on things like `Options()` and highly effective in places like ^

reporting/reporting.go
christineyen: >  a dynamic function representing the number of events processed since the last message would be kinda neat.

Hrmmmm. I hear what you're saying... except `AddDynamicField` gets called on each event we want to send, yeah? So wouldn't it just always be 1?

Is **this** machine's hostname useful or interesting? What about CPU utilization or i/o or something? Any votes for items there? :)

christineyen: > I had a similar issue and never found a lightweight implementation

[:'( via SO](https://stackoverflow.com/a/17783687)

And something like pprof probably isn't something we want to shove in by default. =\ Mrrr.

reporting/reporting.go
```diff
@@ -0,0 +1,137 @@
...
+	contextKeySampler = "sampler"
+	reportingSuffix   = "-reports"
+)
+
+func NewContext(ctx context.Context) context.Context {
+	builder := libhoney.NewBuilder()
+	builder.Dataset += reportingSuffix
+	sampler := &dynsampler.PerKeyThroughput{
+		ClearFrequencySec:      1,
+		PerKeyThroughputPerSec: 1000,
```
christineyen: Psst, a sanity check on these values would be great.

I wish I could configure dynsampler to check after the first second and then adjust every 30, but that seems... not in scope here.

reporting/reporting.go
maplebed: Yes, I think this host's name is a good thing to add for when you're trying to diagnose anything running on a cluster. (I know at least a few folks use honeytail in this way.)

reporting/reporting.go
maplebed: Per key throughput of 1000 is *super* high.

christineyen:
Ping! Rebased off master and made sure to add support for the new postgres/regex parsers.

christineyen:
WHAT is the point of having this two-year-old approved PR floating around. Closing.

PR #76, opened 2017-11-10
Fix race conditions in parser unit tests

emfree:
Tests for the keyval parser were using a `NoopLineParser` that wasn't safe for
concurrent use. But it isn't really necessary in those tests, so just
initialize and use a normal `Parser` instead.

Note: this doesn't fix all the problems with the tests. The unit tests in
`tail_test.go` still have a data race, and the flaky tests we see in travis
probably aren't all fixed by this. But it's a start.


maplebed:
💚

PR #78, opened 2017-11-10
Support multiple log line regexes passed to regex parser

rfong:
Now supports multiple line regexes. Will parse line using first `line_regex` that matches the line. Updated unit tests to match. Will send a `www` pull request updating the relevant docs in a sec.

Would love more edge case test ideas :)

Example usage:
```
go install . && \
honeytail -p regex -k $WRITEKEY --backfill \
  -f PATH/system.log \
  --dataset 'MY_TEST_SET' \
  --regex.line_regex="(?P<time>\d{2}:\d{2}:\d{2}) (?P<computer_name>[a-zA-Z\-]+)" \
  --regex.line_regex="(?P<foo>bar)" \
  --regex.timefield="time" --regex.time_format="%H:%M:%S"
```


parsers/regex/regex.go
```diff
@@ -20,11 +20,10 @@ import (
 )
 
 type Options struct {
-	LineRegex       string `long:"line_regex" description:"a regular expression with named capture groups representing the fields you want parsed (RE2 syntax)"`
-	TimeFieldName   string `long:"timefield" description:"Name of the field that contains a timestamp"`
-	TimeFieldFormat string `long:"time_format" description:"Timestamp format to use (strftime and Golang time.Parse supported)"`
-
-	NumParsers int `hidden:"true" description:"number of regex parsers to spin up"`
+	LineRegex       []string `long:"line_regex" description:"regular expression with named capture groups representing the fields you want parsed (RE2 syntax); can enter multiple regexes to match (--regex.line_regex=\"(?P<foo>re)\" --regex.line_regex=\"(?P<bar>...)\")"`
```
maplebed: nit: let's pluralize `LineRegexes` for easier reading

parsers/regex/regex.go
```diff
@@ -34,74 +33,110 @@ type Parser struct {
...
-	match := p.lineRegex.FindAllStringSubmatch(line, -1)
-	if match == nil || len(match) == 0 {
-		logrus.WithFields(logrus.Fields{
-			"line": line,
-		}).Debug("No matches for regex log line")
-		return parsed, nil
+// RegexLineParser factory
+func NewRegexLineParser(regexStrs []string) (*RegexLineParser, error) {
+	lineRegexes, err := ParseLineRegexes(regexStrs)
+	fmt.Println("lineRegexes", lineRegexes)
```
maplebed: leftover debugging?

parsers/regex/regex.go
```diff
@@ -34,74 +33,110 @@ type Parser struct {
...
-		if i != 0 && i < len(firstMatch) {
-			parsed[name] = firstMatch[i]
+func (p *RegexLineParser) ParseLine(line string) (map[string]interface{}, error) {
+	for _, lineRegex := range p.lineRegexes {
+		parsed := make(map[string]interface{})
+		match := lineRegex.FindAllStringSubmatch(line, -1)
+		if match == nil || len(match) == 0 {
+			logrus.WithFields(logrus.Fields{
+				"line": line,
+			}).Debug("No matches for regex log line")
```
maplebed: if you still want to spit out this debugging line (which might now happen multiple times) let's add the regex it's using to try and match so it's not confusing to see the same line listed as "no matches" several times.

parsers/regex/regex.go
```diff
@@ -34,74 +33,110 @@ type Parser struct {
...
+				parsed[name] = firstMatch[i]
+			}
+		}
+		logrus.WithFields(logrus.Fields{
+			"parsed":    parsed,
+			"line":      line,
+			"lineRegex": lineRegex,
+		}).Debug("Regex parsing log line")
+
+		return parsed, nil
```
maplebed: I'd like a comment somewhere (maybe the overall docs?) saying something like "first regex to match the line wins, so make sure to list your regexes in most-specific to least-specific order"

parsers/regex/regex_test.go
```diff
@@ -136,25 +159,52 @@ var tlms = []testLineMap{
...
+		// Multi-regex parsing
+		lineRegexes: []string{
+			`\[(?P<word1>\w+)\]`,
+			`(?P<dummy>banana)`,
+		},
+		input: `apple banana orange`, // Should match to second regex
+		expected: map[string]interface{}{
+			"dummy": "banana",
+		},
+	},
```
maplebed: could we add a test for a line that matches none of the listed regexes and see it fall through?

parsers/regex/regex_test.go
```diff
@@ -136,25 +159,52 @@ var tlms = []testLineMap{
...
-		lineParser: &RegexLineParser{regexp.MustCompile(``)},
-	}
 	for _, tlm := range tlms {
-		p.lineParser = &RegexLineParser{regexp.MustCompile(tlm.lineRegex)}
+		p := &Parser{}
+		err := p.Init(&Options{
+			NumParsers: 5,
+			LineRegex:  tlm.lineRegexes,
+		})
+		if err != nil {
```
maplebed: totally optional but `leash.go` uses the `stretchr/testify/assert` library, which has `assert.NoError(...)` for exactly this kind of test - that the function did not return an error. https://godoc.org/github.com/stretchr/testify/assert#NoError

PR #79, opened 2017-11-11
When consuming lines, only trim whitespace in parsers (fixes postgres)

emfree:
This is an issue that currently isn't caught by unit tests. In the postgres
parser, we rely on the existence of leading whitespace to distinguish lines
that are the continuation of a SQL statement, e.g.:

```
2017-11-07 01:43:39 UTC [3542-7] postgres@test LOG:  duration: 15.577 ms  statement: SELECT * FROM test
		WHERE id=1;
```

from lines that are a new statement.

But in `tail.go`, we were stripping leading and trailing whitespace. Do you
remember what the scenarios were that we needed that for? For the moment, I've
been conservative and lifted the `strings.TrimSpace` call into all the parsers
except the postgres one. Not sure if this is the best approach. It does mean
that this won't accidentally break anything.

### DIFF
--- a/parsers/arangodb/arangodb.go
+++ b/parsers/arangodb/arangodb.go
@@ -173,7 +173,7 @@ func (p *Parser) ProcessLines(lines <-chan string, send chan<- event.Event, pref
 		wg.Add(1)
 		go func() {
 			for line := range lines {
-
+				line = strings.TrimSpace(line)
 				// take care of any headers on the line
 				var prefixFields map[string]string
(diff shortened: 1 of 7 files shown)


nathanleclaire:
I like this solution -- while it feels a bit awkward to have `strings.TrimSpace` in so many places, it's ultimately the parser's responsibility to decide where an event starts or ends in the file.

nathanleclaire:
LGTM

PR #84, opened 2017-12-22
Parse prepared statements in postgres logs

emfree:
PostgreSQL prepared statements (i.e., statements with bind parameters
rather than inline parameters) look like this in the logs:

```
2017-12-22 22:14:10 UTC:10.2.31.141(51017):honeycomb@sbtest:[2035]:LOG:  duration: 0.065 ms  bind sbstmt-513671205-1941546510: SELECT c FROM sbtest1 WHERE id BETWEEN $1 AND $2 ORDER BY c
2017-12-22 22:14:10 UTC:10.2.31.141(51017):honeycomb@sbtest:[2035]:DETAIL:  parameters: $1 = '5043', $2 = '5142'
2017-12-22 22:14:10 UTC:10.2.31.141(51017):honeycomb@sbtest:[2035]:LOG:  duration: 0.218 ms  execute sbstmt-513671205-1941546510: SELECT c FROM sbtest1 WHERE id BETWEEN $1 AND $2 ORDER BY c
```

So there are three log lines for each query: "bind", "parameters" and
"execute". Currently, we fail to parse these at all. As an interim
measure, just parse the "execute" line. It would be nice to parse the
parameters as well, but that would require more logic to glue the lines
together. (People often don't want to send the raw parameters anyways,
out of concern for sensitive data.)

### DIFF
--- a/parsers/postgresql/postgresql.go
+++ b/parsers/postgresql/postgresql.go
@@ -68,7 +68,7 @@ const (
 	timestampRe   = `\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}[.0-9]* [A-Z]+`
 	defaultPrefix = "%t [%p-%l] %u@%d"
 	// Regex string that matches the header of a slow query log line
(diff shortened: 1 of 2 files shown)


maplebed:
Excellent, thank you. LGTM.

PR #86, opened 2018-02-06
Fix log rotation handling

emfree:
There are two ways to rotate a file: 
1. By renaming `foo.log` to `foo.log.1` and then recreating `foo.log`.
2. By copying `foo.log` to `foo.log.1` and then truncating `foo.log`. 

A [bug](https://github.com/hpcloud/tail/pull/115) in the `hpcloud/tail` library meant that honeytail could fail to keep tailing a file after a rename/reopen, unless you were using the `--tail.poll` option. Thanks @adihoney for finding this.

The rather elaborate included test checks all different configurations of log rotation and polling vs inotifying, and shows that updating `hpcloud/tail` should fix the issue. Note: there still seems to be a bug where honeytail loses the first line of a file after log rotation if you're using copy/truncate rotation. I will shave that yak later.

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -76,28 +76,28 @@
 		},
 		{
 			"ImportPath": "github.com/hpcloud/tail",
-			"Comment": "v1.0.0",
-			"Rev": "a30252cb686a21eb2d0b98132633053ec2f7f1e5"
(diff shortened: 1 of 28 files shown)


adihoney:
LGTM!

PR #90, opened 2018-02-12
add reference of keyval/logfmt parser to README

igorwwwwwwwwwwwwwwwwwwww:
We were super psyched to learn that honeytail supports our logfmt logs out of the box, and would like to make this information easier to find.

Honeytail is bringing us a lot of joy! 💜


emfree:
Thank you! :sparkling_heart:

PR #91, opened 2018-02-15
parsers/postgresql: ensure prefix matching actually matches the beginning of the string

igorwwwwwwwwwwwwwwwwwwww:
ran into this while trying to consume heroku postgres logs with honeytail.

this matching causes issues in `parsePrefix`, because it assumes the match starts at the beginning of the line. so when it doesn't, the substring slicing may end up not stripping the whole prefix, causing issues with the rest of the parsing down the line.


emfree:
Awesome, thanks for the PR!

PR #97, opened 2018-03-12
Add CSV parser

tredman:
It's 2018 and some things still log in CSV (looks at MariaDB audit plugin).

### DIFF
--- a/parsers/csv/csv.go
+++ b/parsers/csv/csv.go
@@ -1,0 +1,155 @@
+package csv
+
+import (
+	"encoding/csv"
+	"errors"
+	"strconv"
+	"strings"
+	"sync"
+
+	"github.com/Sirupsen/logrus"
+
+	"github.com/honeycombio/honeytail/event"
+	"github.com/honeycombio/honeytail/httime"
+	"github.com/honeycombio/honeytail/parsers"
+)
+
+// Options defines the options relevant to the CSV parser
+type Options struct {
+	Fields           string `long:"fields" description:"Comma separated list of CSV fields, in order."`
+	TimeFieldName    string `long:"timefield" description:"Name of the field that contains a timestamp"`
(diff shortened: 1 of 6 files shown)


parsers/csv/csv.go
```diff
@@ -0,0 +1,152 @@
...
+	fields []string
+}
+
+// NewCSVLineParser factory
+func NewCSVLineParser(fieldsString string) (*CSVLineParser, error) {
+	// Is building a reader for every single line a good idea?
+	// Potential for future optimization here
+	reader := strings.NewReader(fieldsString)
+	csvReader := csv.NewReader(reader)
+	csvReader.TrimLeadingSpace = true
```
emfree: We set `TrimLeadingSpace = true` here, but not on the `csvReader` down in `ParseLine()` -- probably best to be consistent.

parsers/csv/csv.go
```diff
@@ -0,0 +1,152 @@
...
+	if err != nil && err != io.EOF {
+		logrus.WithError(err).WithField("line", line).
+			Warn("failed to parse line")
+		return nil, err
+	}
+
+	lenValues := len(values)
+	lenFields := len(p.fields)
+
+	if lenValues != lenFields {
```
emfree: I think it might actually be safer to require the number of fields in each line to match the number of fields in `p.fields`. Otherwise it'd be pretty easy to typo `--csv.fields` and have a silent mismatch between the field names and values in honeycomb. Looks like we can require a specific number of fields by setting `csvReader.FieldsPerRecord`. Thoughts?

tredman: Yeah I can see that. I was thinking of how we'd want to keep logging (and just warn) even if the number of fields increased or was variable for some reason, but that actually seems like a bad pattern to get into since ordering of fields is important in CSV. I'll change this to error and not send to honeycomb in this scenario.

parsers/csv/csv.go
tredman: I'm going to make this optional for both fields and values, defaulting to false.

PR #101, opened 2018-04-26
Add --filter-files option to honeytail

tredman:
Got a customer request to add this capabilility about a month ago. Seems straightforward enough. I tested this locally in addition to the supplied test.


tredman:
From our asana task on this:

> Each container on DC/OS contains a specific directory with logs, state and control file. To collect only the logs, they need to tell Honeytail to ignore the state and control files.

>  We have a directory log app when many logs but I would like to have a except for any log name contains "state" like: app_20180214_state.log

PR #102, opened 2018-05-07
Implement --deterministic_sampler

nathanleclaire:
Current method is based on ~~[Jump Hash](https://github.com/renstrom/go-jump-consistent-hash)~~ SHA1. SHA1 is a well-known hash with desirable properties (even distribution) in the standard library. If it proves too resource intensive, we can modify it, but we'll start here.

### DIFF
--- a/leash.go
+++ b/leash.go
@@ -18,7 +18,7 @@ import (
 	"time"
 
 	"github.com/Sirupsen/logrus"
-	"github.com/honeycombio/dynsampler-go"
+	dynsampler "github.com/honeycombio/dynsampler-go"
 	"github.com/honeycombio/libhoney-go"
 	"github.com/honeycombio/urlshaper"
 
@@ -34,6 +34,7 @@ import (
 	"github.com/honeycombio/honeytail/parsers/postgresql"
 	"github.com/honeycombio/honeytail/parsers/regex"
 	"github.com/honeycombio/honeytail/parsers/syslog"
+	"github.com/honeycombio/honeytail/sample"
 	"github.com/honeycombio/honeytail/tail"
 )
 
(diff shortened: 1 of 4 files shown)


nathanleclaire:
a little diagram how this is intended to work in case it's not clear:

image

(sample rate would be 4 in this diagram)

every deterministic sampling field val will get hashed (SHA1) - the number of possible values in SHA1 hash in Golang is (2 ** (8*20))-1 (hence the `big.Int`). Everything up to (this number / sample rate) will be sampled, thus giving us about 1/N events while also sampling the same events every time.

leash.go
```diff
@@ -352,18 +363,32 @@ func modifyEventContents(toBeSent chan event.Event, options GlobalOptions) chan
...
+								Error("Field to deterministically sample on does not exist in event, leaving it to random chance")
+							if rand.Intn(int(options.SampleRate)) != 0 {
+								ev.SampleRate = -1
+							}
+						} else {
+							if !deterministicSampler.Sample(sampleKey) {
+								ev.SampleRate = -1
+							}
+						}
+						logrus.WithField("sampleKey", sampleKey).WithField("bucket", deterministicSampler.Sample(sampleKey)).Info("hash")
```
maplebed: is this leftover debugging?

nathanleclaire: ah, seems to be, yes

sample/deterministic_sampler_test.go
```diff
@@ -0,0 +1,84 @@
...
+
+func randomRequestID() string {
+	// create request ID roughly resembling something you would get from
+	// AWS ALB, e.g.,
+	//
+	// 1-5ababc0a-4df707925c1681932ea22a20
+	//
+	// The AWS docs say the middle bit is "time in seconds since epoch",
+	// (implying base 10) but the above represents an actual Root= ID from
+	// an ALB access log, so... yeah.
```
adihoney: 5ababc0a is hex for 1522187274 which is 2018-03-27, so the docs aren't wrong, just confusing. :)

nathanleclaire: ah, i was wondering about that. good call out

PR #103, opened 2018-05-10
use uint32 rather than big.Int in DeterministicSampler

adihoney:
Since sample rates range from 1 up to around a million (in the most extreme case), we can avoid some allocations by just considering the first 32 bits of the hash.

Also add a known answer test to confirm that this doesn't change the behavior (the test passes before and after the big.Int -> uint32 change).


nathanleclaire:
LGTM

PR #105, opened 2018-07-05
Add --rebase_time argument

tredman:
When set, timestamps are "rebased" onto the current time. The last event (chronologically) in the log file(s) is set to the current time. Event timestamps before this are updated relative to this time, with the time delta preserved.

For example, if now is Wed Jul 4 12:00:00 -0800 PST 2018, and the events start and end at Wed Jul 3 15:04:05 -0800 PST 2018 and Wed Jul 3 12:00:05 -0800 PST 2018, respectively, the events will be updated so that they appear to have started and ended at Wed Jul 4 08:56:00 -0800 PST 2018 and Wed Jul 4 12:00:00 -0800 PST 2018.

The goal is to allow you to input old data but look at it as though it happened recently.

### DIFF
--- a/leash.go
+++ b/leash.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bufio"
 	"context"
 	"crypto/sha256"
 	"encoding/json"
@@ -9,6 +10,7 @@ import (
 	"math/rand"
 	"os"
 	"os/signal"
+	"path/filepath"
 	"regexp"
 	"sort"
 	"strconv"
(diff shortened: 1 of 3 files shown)


leash.go
```diff
@@ -605,3 +621,93 @@ func logStats(stats *responseStats, interval uint) {
...
+
+func getEndLine(file string) string {
+	handle, err := os.Open(file)
+	if err != nil {
+		logrus.WithError(err).WithField("file", file).
+			Fatal("unable to open file")
+	}
+	defer handle.Close()
+	// we use a scanner to read to the last line
+	// not the most efficient
```
maplebed: this is going to be super bad on a multi-gigabyte file. Can we open the file at the end then read back a reasonable amount and start scanning there instead?

PR #106, opened 2018-07-16
Bump Go version and enable OSX builds

nathanleclaire:
Since we will have the sample JSON file as part of the Gatekeeper tour I think we will want to revisit this.

I kind of wish we could get a Windows build together as well...


nathanleclaire:
yep

PR #107, opened 2018-07-17
add presampled option for ingesting already-sampled data

maplebed:
For the gatekeeper demo, we have a dataset that has a sample rate embedded in the JSON already. We need a flag to say that we should pay attention to the sample rate that's already happened rather than applying sampling at honeytail.

### DIFF
--- a/leash.go
+++ b/leash.go
@@ -377,28 +377,45 @@ func modifyEventContents(toBeSent chan event.Event, options GlobalOptions) chan 
 					for k, v := range parsedAddFields {
 						ev.Data[k] = v
 					}
-					ev.SampleRate = int(options.SampleRate)
-					if dynamicSampler != nil {
-						key := makeDynsampleKey(&ev, options)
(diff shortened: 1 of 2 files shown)


leash.go
```diff
@@ -377,28 +377,45 @@ func modifyEventContents(toBeSent chan event.Event, options GlobalOptions) chan
...
-							ev.SampleRate = sr
+					// get presampled field if it exists
+					if options.PreSampledField != "" {
+						var presampledRate int
+						if psr, ok := ev.Data[options.PreSampledField]; ok {
+							switch psr := psr.(type) {
+							case float64:
+								presampledRate = int(psr)
+							case string:
+								if val, err := strconv.Atoi(psr); err == nil {
```
christineyen: are we doing anything with this error?

maplebed: probably should but honeytail doesn't have great support for communicating this sort of thing. :/

PR #109, opened 2018-07-19
[time] change logic for handling fractional epoch time to deal with fractional precision

maplebed:
Neither go nor strftime deal with times like 1532024683.0938044 (representing Thursday, July 19, 2018 6:24:43 PM and some fractional second). But this code isn't really doing strftime, it's trying to detect whether it's a unix time first. Also, there's no reason to pin it to exactly 3 allowed digits after the second when we should reasonably accept any fraction, regardless of precision.

### DIFF
--- a/httime/httime.go
+++ b/httime/httime.go
@@ -13,6 +13,7 @@ const (
 	StrftimeChar        = "%"
 	UnixTimestampFmt    = "%s(%L)?"
 	UnixTimestampFmtAlt = "%s(.%L)?"
+	UnixTimestampFmtTxt = "unixepoch"
 )
 
 // TODO UnixTimestampFmt is not a regex - it shouldn't look like one. Also even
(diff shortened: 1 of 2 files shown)


httime/httime.go
```diff
@@ -10,10 +10,24 @@ import (
...
+	UnixTimestampFmtAlt = "%s(.%L)?"
+	UnixTimestampFmtTxt = "unixepoch"
 )
 
+// TODO UnixTimestampFmt is not a regex - it shouldn't look like one. Also even
+// if it were a regex it would be wrong because it's expecting a dot between the
+// seconds and fractional seconds, eg `%s(.%L)?`. But just changing it would
+// break anybody that is using the magical  string `%s(%L)?` as their timestamp
+// in order to enable this magical behavior. Is anybody doing that? Who can
+// know? Introducing UnixTimestampFmtAlt gets it closer to what it's actually
```
nathanleclaire: > Who can know?

This comment is headed in a dangerously existential (well, epistemological really) direction :D

nathanleclaire:
What's the motivation for this? Did someone using honeytail or httime hit issues with precision ?

nathanleclaire:
LGTM though - just curious

maplebed:
Trying to process a log that is using fractional epoch timestamps - it's not reliably millisecond or microsecond or nanosecond, it's just seconds + fraction eg 1532024683.0938044

PR #111, opened 2018-09-08
Add option to unescape JSON fields encoded as strings

nathanleclaire:
This will allow JSON data stored as a string in files to be parsed as a proper object by Honeytail and send as non-escaped JSON to the API.


nathanleclaire:
🙏

PR #112, opened 2018-09-20
[WIP] Have Honeytail pick up on newly created files when using globs

nathanleclaire:
Signed-off-by: Nathan LeClaire <nathan.leclaire@gmail.com>

PR #113, opened 2018-09-28
Use chan chan string instead of returning chans

nathanleclaire:
This allows for new `lines` channels to be sent down the pipe.

Successor/competitor to: https://github.com/honeycombio/honeytail/pull/112

Still rough and WIP, just want to start posting in the PR.

### DIFF
--- a/leash.go
+++ b/leash.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"crypto/sha256"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -24,6 +25,7 @@ import (
 	dynsampler "github.com/honeycombio/dynsampler-go"
 	"github.com/honeycombio/libhoney-go"
 	"github.com/honeycombio/urlshaper"
+	fsnotify "gopkg.in/fsnotify.v1"
 
 	"github.com/honeycombio/honeytail/event"
 	"github.com/honeycombio/honeytail/parsers"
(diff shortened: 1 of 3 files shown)


nathanleclaire:
Probably will put `removeStateFiles` back in `tail`, looking at it now I'm not sure I'm a fan of having it in `leash.go` instead of just exporting it.

nathanleclaire:
TODO:
- [ ] finish fixing unit tests
- [ ] vendor fsNotify dep

tail/tail_test.go
```diff
@@ -115,43 +121,21 @@ func TestGetEntries(t *testing.T) {
...
 		jsonLines[i] = make([]string, 3)
 		for j := 0; j < 3; j++ {
 			jsonLines[i][j] = fmt.Sprintf("{\"a\":%d}", i)
 		}
 
 		filename := filenameRoot + fmt.Sprint(i)
-		conf.Paths[i] = filename
+		fileChan <- filename
 		ts.writeFile(t, filename, strings.Join(jsonLines[i], "\n"))
-	}
```
nathanleclaire: Not crazy about taking the below chunk of the test out, but it was definitely oriented towards the old model of returning a list of `chan string` (in this case, an empty one) instead of taking in a `chan chan string` that gets written to. Any ideas or do you think it's OK to just remove?

leash.go
```diff
@@ -41,15 +43,139 @@ import (
...
+		}
+	}
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case fsEvent := <-filewatcher.Events:
+			logrus.Warn(fsEvent)
+			if fsEvent.Op == fsnotify.Create {
```
tredman: do we need to handle removals or does that get handled by the tailer detecting EOF/read error?

tredman:
@nathanleclaire This LGTM but can you address the test failures? :)

PR #120, opened 2018-10-25
Add support for parsing floating point timestamps, like nginx $msec

ngauthier:
The only way to get more than second precision in nginx is to use the $msec variable, which has millisecond precision. Honeytail recognizes it as a floating point number and converts it, but then fails to recognize that as a valid timestamp because it's not a string.

This PR allows honeytail to automatically recognize the msec field (to skip some searching during parsing) and also to recognize a floating point timestamp as a timestamp in seconds.

Unfortunately, without a refactor, we can't retain the originally parse precision, we have to go to string then back to float and then break that float into a second and nanosecond part, which means we lose precision somewhere around the microsecond level. I am totally cool with that, just letting you know here that there are rounding issues.

All in all, this means that by including $msec in your log line *instead of `time_local` or `time_iso8601`* (because those are tried first) you can get millisecond precision in your starting timestamps. This is especially useful for small traces.

Thanks,
Nick

### DIFF
--- a/httime/httime.go
+++ b/httime/httime.go
@@ -40,6 +40,7 @@ var (
 		"timestamp", "Timestamp", "TimeStamp",
 		"date", "Date",
 		"datetime", "Datetime", "DateTime",
+		"msec",
 	}
 	// reference: http://man7.org/linux/man-pages/man3/strftime.3.html
 	convertMapping = map[string]string{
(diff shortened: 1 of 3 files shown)


tredman:
LGTM. Also not a fan of converting to string and back, but we can address this without too much work if someone raises an concern regarding precision. Thanks for the PR!

tredman:
btw, I need to wait for the build to finish and then publish a new version to the website. Will comment here when that is done.

ngauthier:
Great, thanks so much!

tredman:
Version 1.730 is up. Install docs are updated too: https://docs.honeycomb.io/getting-data-in/honeytail/

ngauthier:
Thanks! Working for me now :)

PR #132, opened 2019-03-21
If --debug_stdout is set, we don't need a write key

ismith:
Note: the change in the vendor'd honeycombio/libhoney-go is not compatible with upstream api changes, and will need to be revisited, probably soon

### DIFF
--- a/main.go
+++ b/main.go
@@ -181,12 +181,16 @@ func main() {
 	addParserDefaultOptions(&options)
 	sanityCheckOptions(&options)
 
-	if _, err := libhoney.VerifyWriteKey(libhoney.Config{
-		APIHost:  options.APIHost,
-		WriteKey: options.Reqs.WriteKey,
-	}); err != nil {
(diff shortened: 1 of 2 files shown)


ismith:
Without this, you can't run honeytail without a valid -k, even in --debug_stdout

ismith:
@maplebed sounds like you have a plan for proceeding? I'm happy to help with changes, but updating to current libhoney-go felt like a separate (logical) bit of work from the honeytail-local exception.

ismith:
@maplebed I went to look at libhoney-go and what'd be necessary there and ... there was already a comment to the effect that maybe those checks should be looser :joy: lmk what you think.

ismith:
(And that change to libhoney-go works with my test above if I rm the vendor'd libhoney-go, so updating this package's libhoney to be more current should just work.)

ismith:
@maplebed do this and https://github.com/honeycombio/libhoney-go/pull/48 look like the right things?

PR #133, opened 2019-03-22
Multiline json

ismith:
Also tested manually, using 
`cat multiline.log | honeytail --dataset some-data --parser json  --debug_stdout --debug  -f -` (note: `-k` is missing, my branch was rebased on https://github.com/honeycombio/honeytail/pull/132)
against a file multiline.log:
```json
"unfinished log item": 100
}
{
  "foo": 1,
  "bar": 2
}

{ "bar": 3
}

{

  "foo": "{"
}

{
 "deadkey": 1

```

Results in two errors (for the incomplete objects at beginning and end) and three successfully parsed events.

### DIFF
--- a/parsers/htjson/htjson.go
+++ b/parsers/htjson/htjson.go
@@ -73,31 +73,25 @@ func (p *Parser) ProcessLines(lines <-chan string, send chan<- event.Event, pref
 	// figure out beginning
 	for line := range lines {
 		line = strings.TrimSpace(line)
-		if strings.HasPrefix(line, "{") {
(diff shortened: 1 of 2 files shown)


ismith:
Not sure why this passed on Linux but not OSX ...

ismith:
This used to fail in CI due to a timeout, see dad824c for the fix.

It's now failing in the `watch` step with error `buildevents - Error detected: circle token required to poll the API`, which I don't think is related to my code, maybe just a buildevents/circle config problem?

PR #150, opened 2020-03-17
Add duration conversion

SeanHood:
Certain applications (httpd in my case) output their duration in microseconds
however honeycomb assumes time is in milliseconds. This PR adds two configuration
options to help with this conversion before the events are sent.

`DurationField / --duration_field`: This option is where you set the field you
wish to convert.

`DurationUnit / --duration_unit`: This option is where you set the unit of the
incoming duration. Accepted options: `s` (seconds) and `us` (microseconds).

For the `nginx` parser, we default to `request_time` so we only have to configure `DurationUnit` in most cases.

Discussion: https://honeycombpollinators.slack.com/archives/C4T3A32CX/p1583938457301600

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -9,3 +9,4 @@ Ian Wilkes
 Justin Hennessy
 Max Neunhoeffer
 Rick Branson
+Sean Hood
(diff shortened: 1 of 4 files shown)


SeanHood:
I feel there's maybe some changes to be made to cleanup the duplication of code, however I'm not a Go developer so unsure how to clean this up.

MikeGoldsmith:
We’re going to close this for now as there has been no recent activity. Feel free to re-open if it’s still relevant.

PR #156, opened 2020-06-01
[honeytail] Fix builds so we can release v1.1.2

dsoo:
### Summary
This change fixes our automated builds so that we can deploy artifacts to Github Releases and S3 automatically.
* Adds CHANGELOG.md to track changes.

There are no changes to the `honeytail` binary, just changes to our build and deployment tooling.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -125,8 +125,8 @@ jobs:
                 aws-secret-access-key: AWS_SECRET_ACCESS_KEY
                 aws-region: AWS_REGION
             - buildevents/berun:
-                bename: sync artifacts to s3
-                becommand: aws s3 sync ~/artifacts s3://honeycomb-builds/honeycombio/honeytail/$VERSION/
+                bename: sync_s3_artifacts
+                becommand: aws s3 sync ~/artifacts s3://honeycomb-builds/honeycombio/honeytail/${VERSION}/
 
 workflows:
   build:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -98,7 +98,7 @@ jobs:
             - store_artifacts:
                 path: ~/artifacts
 
-  publish:
+  publish_github:
     docker:
       - image: cibuilds/github:0.13.0
     steps:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -79,34 +79,30 @@ jobs:
...
-                becommand: ./build-pkg.sh -m amd64 -v "${CIRCLE_TAG}" -t deb
+                becommand: |
+                  ./build-pkg.sh -m amd64 -v "${CIRCLE_TAG}" -t deb
+                  mv *.deb ~/artifacts
             - buildevents/berun:
                 bename: build_deb_arm64
-                becommand: ./build-pkg.sh -m arm64 -v "${CIRCLE_TAG}" -t deb
+                becommand: |
+                  ./build-pkg.sh -m arm64 -v "${CIRCLE_TAG}" -t deb
+                  mv *.deb ~/artifacts
```
maplebed, line 92: I don't think `becommand` can take multiple commands in the same way the regular `run` command does - I think `becommand` is restricted to a single thing executable by a shell. If this actually works then my memory is wrong and carry on!

maplebed: Looking at the build output, I think the result of this syntax is that it runs the `./build-pkg.sh` within the scope of the span representing `build_deb_arm64` but the execution of the `mv` command is outside of it. So while both commands might both be executed, only the former is represented by the span. This makes the code here misleading in terms of what the trace output will represent. :/ Not quite as bad as I feared (that it wouldn't run at all) but we shouldn't land it like this.

dsoo: Cool, I will fix it. I'd note that we do it this way in at least one other repo: https://github.com/honeycombio/basenji/blob/master/.circleci/config.yml
Do you think it's possible to fix `becommand` to handle this case properly?

dsoo:
Hang on, don't re-review this yet - I'm making sure that the upload and nginx redirects for the docs site will work.

dsoo:
@maplebed - cleaned this up, in the context of further investigation of the rest of our build pipeline. There are other PRs inbound for related changes elsewhere.

PR #167, opened 2020-12-16
Use public ENV context in circleci

vreynolds:
* currently secrets are pulled from the project env variables in CircleCI
* using a shared context consolidates secrets storage, makes for easier secret updates
* majority of other repos already use the shared context, looks like this was set up
a while back before the context was created
* secret values are unchanged from what's currently in the project env variables

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -156,7 +156,7 @@ workflows:
             tags:
               only: /.*/
       - publish_github:
-          context: Honeycomb Secrets
+          context: Honeycomb Secrets for Public Repos
           requires:
             - build
           filters:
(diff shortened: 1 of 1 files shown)


paulosman (approved):

vreynolds:
Each job uses the [buildevents orb](https://github.com/honeycombio/buildevents-orb/), which requires the buildevents env variables. Adding the context in each job is equivalent to what happens today with project env variables. 

### Considerations

* Removing buildevents -- it sounds that in the past having buildevents caused forked PRs not to build. This is not ideal, but I think this is an issues with buildevents/buildevents orb. It should allow the build to run, even if there's no secrets.


@paulosman additional commit adds context to every job

paulosman (approved):

PR #173, opened 2021-02-05
Update dependabot and remove buildevents from circle config

paulosman:
Circle changes should allow project to build on PRs from dependabot and others.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -99,8 +99,6 @@ workflows:
     jobs:
       - build:
           context: Honeycomb Secrets for Public Repos
-          requires:
-            - test
           filters:
             tags:
               only: /.*/
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,7 +22,6 @@ commands:
       - run: |
           GOOS=<< parameters.os >> \
           GOARCH=<< parameters.arch >> \
-          buildevents cmd $CIRCLE_WORKFLOW_ID $BUILDEVENTS_SPAN_ID go_build -- \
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -1,15 +1,9 @@
 version: 2.1
 
 orbs:
-  buildevents: honeycombio/buildevents@0.2.6
   aws-cli: circleci/aws-cli@0.1.13
+  docekr: circleci/docker@1.3.0
```
robbkidd, line 5: docekr?

```suggestion
  docker: circleci/docker@1.3.0
```

robbkidd (changes requested):

.circleci/config.yml
paulosman: @robbkidd yours is better

.circleci/config.yml
```diff
@@ -28,137 +22,94 @@ commands:
...
+      - image: circleci/golang:1.13
     steps:
-      - buildevents/with_job_span:
-          steps:
-            - checkout
-            - buildevents/berun:
-                bename: go_test
-                becommand: go test --timeout 10s -v ./...
+      - checkout
+      - run: go test --timeout 10s -v ./...
```
robbkidd, line 35: If I'm reading this correctly, the `build` job is now responsible for running the tests and there is no `test` job any more, right? I'd expect this job to be named `test`. 😄

paulosman: You are making sense.

robbkidd (approved):
YES

PR #175, opened 2021-02-08
Publishing to s3 and gh requires that the packages are built first

paulosman:
Are unit tests for Circle configs a thing? :)


vreynolds (approved):
📦

PR #179, opened 2021-03-02
remove unnecessary context from test job

vreynolds:


robbkidd (approved):
Get out of here, context!

Get out!

PR #180, opened 2021-03-02
add the label manager GitHub action

robbkidd:


robbkidd (commented):
I'm reviewing myself for this administrivia PR. Labels are applied. 👍

PR #181, opened 2021-03-03
tell Dependabot to use our maintenance label

robbkidd:
Use our labeling scheme instead of calling it "dependencies"


vreynolds (approved):
🔥

PR #191, opened 2021-04-20
Add option to include file path hash to statefile directory path

MikeGoldsmith:
**Summary**

When using a file glob that returns multiple files that share the same filename, a single state file is created with increased write contention. Instead, a state file should be created per file that is being tailed.

This change adds a new tail option to generate a hash for each filename directory that is appended to the state file and ensures each tailed file has a unique state file. State files are generated with the format `filename.leach.state-{hash}`.

### DIFF
--- a/tail/tail.go
+++ b/tail/tail.go
@@ -419,7 +419,7 @@ func getStateFile(conf Config, filename string, numFiles int) string {
 
 	var stateFileName string
 	if conf.Options.HashStateFileDirPaths {
-		// generate hash based on filepath, in format 'filename.leach.state-{hash}'
+		// generate hash based on filepath, in format 'filename.leash.state-{hash}'
 		stateFileName = strings.TrimSuffix(filepath.Base(filename), ".log") + ".leash.state" + fmt.Sprintf("-%x", sha1.Sum([]byte(confStateFile)))
 	} else {
 		stateFileName = strings.TrimSuffix(filepath.Base(filename), ".log") + ".leash.state"
--- a/tail/tail.go
+++ b/tail/tail.go
@@ -7,6 +7,7 @@ package tail
 import (
 	"bufio"
 	"context"
+	"crypto/sha1"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -16,8 +17,8 @@ import (
 	"strings"
 	"time"
 
-	"github.com/sirupsen/logrus"
 	"github.com/hpcloud/tail"
+	"github.com/sirupsen/logrus"
 	"golang.org/x/sys/unix"
 )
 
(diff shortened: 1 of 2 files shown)


tail/tail.go
```diff
@@ -415,7 +417,13 @@ func getStateFile(conf Config, filename string, numFiles int) string {
 		}
 	}
 
-	stateFileName := strings.TrimSuffix(filepath.Base(filename), ".log") + ".leash.state"
+	var stateFileName string
+	if conf.Options.HashStateFileDirPaths {
+		// generate hash based on filepath, in format 'filename.leach.state-{hash}'
```
vreynolds, line 422: comment has a typo? leach -> leash

vreynolds (approved):
Looks good. Should we add something in the docs about this option? https://docs.honeycomb.io/getting-data-in/integrations/honeytail/

vreynolds (approved):

robbkidd (approved):
hashes playing tag

netanswer-devs:
I was the one who intially reported this performance issue.
I've installed honeytail 1.4.0 and I still have the issue.
With
`--tail.hash_statefile_paths  --file=/var/www/*/log/daily/beeline_json.log`
A "lsof" gives me

```
honeytail 10207 admin  256r   REG      254,1  49751336    7816077 /var/www/client1/log/daily/beeline_json.log
honeytail 10207 admin  257u   REG      254,1        38     151295 /tmp/beeline_json.leash.state-8c393341f53683eee209137292b19ded935cbedb
honeytail 10207 admin  258r   REG      254,1   8663284   18213825 /var/www/client2/log/daily/beeline_json.log
honeytail 10207 admin  259u   REG      254,1        38     151295 /tmp/beeline_json.leash.state-8c393341f53683eee209137292b19ded935cbedb
honeytail 10207 admin  260r   REG      254,1  10353810   18213829 /var/www/client3/log/daily/beeline_json.log
honeytail 10207 admin  261u   REG      254,1        38     151295 /tmp/beeline_json.leash.state-8c393341f53683eee209137292b19ded935cbedb
honeytail 10207 admin  262r   REG      254,1   4624380   18702927 /var/www/client4/log/daily/beeline_json.log
honeytail 10207 admin  263u   REG      254,1        38     151295 /tmp/beeline_json.leash.state-8c393341f53683eee209137292b19ded935cbedb
honeytail 10207 admin  264r   REG      254,1     92849   21252515 /var/www/client5/log/www1/daily/beeline_json.log
honeytail 10207 admin  265u   REG      254,1        38     151295 /tmp/beeline_json.leash.state-8c393341f53683eee209137292b19ded935cbedb
honeytail 10207 admin  266r   REG      254,1  36479583   21262971 /var/www/client5/log/www2/daily/beeline_json.log
honeytail 10207 admin  267u   REG      254,1        38     151295 /tmp/beeline_json.leash.state-8c393341f53683eee209137292b19ded935cbedb
honeytail 10207 admin  268r   REG      254,1   4070147   21252574 /var/www/client5/log/www4/daily/beeline_json.log
honeytail 10207 admin  269u   REG      254,1        38     151295 /tmp/beeline_json.leash.state-8c393341f53683eee209137292b19ded935cbedb
```
It seems to me that in the initial commit
`stateFileName = strings.TrimSuffix(filepath.Base(filename), ".log") + ".leash.state" + fmt.Sprintf("-%x", sha1.Sum([]byte(confStateFile)))`
should be
`stateFileName = strings.TrimSuffix(filepath.Base(filename), ".log") + ".leash.state" + fmt.Sprintf("-%x", sha1.Sum([]byte(filename)))`
because confStateFile does not change between files so the hash will always be the same

MikeGoldsmith:
@netanswer-devs you're totally right. The hash we're using will always be of the location of the state file (temp folder or custom dir), not the location of the original file that is meant to be tailed 😦 

I'll get this fixed up ASAP.

MikeGoldsmith:
@netanswer-devs we've fixed the hash to be generated based on the tailed file path, not state file path, in [v1.4.1](https://github.com/honeycombio/honeytail/releases/tag/v1.4.1) - let us know if that fixes the issues you've been seeing.

netanswer-devs:
@MikeGoldsmith The issue (same state file opened multiple times) is fixed. 
For ~130 files I was at 225 writes/s and now at 150 writes/s (~10 writes/s without honeytail running) with is good.
I actually have twice that many files (an other 130 files for cron jobs, because it's an other dataset) so about 450/s
Honeytail still do about 75% to 90% of all the writes on the server

We would need honeytail to write to file only if the offset has changed
In my case about 3 600 000 spans/day on 920 000 unique couple (file, second) so 920K writes a day instead of 260*86400, about 4% of the current number of writes (and at most 30 writes/s for 99% of the time)
But our setup with that many websites (and files) might not be the most common one (event then others might not have the same gain with that optimization)...
I've tried a Redis stream on our developement server and our own agent to send the data to your API, seems to be working perfectly.

If you do decide to invest time for that optimization I would gladly test it and report to you, even if we end up keeping the redis stream (having it now I can see other usages which make it interesting for us)

PR #193, opened 2021-04-22
Generate statefile hash based on tailed file location

MikeGoldsmith:
Fixes an issue from #191 where the statefile path is used to generate the hash, instead we should generate the hash based on the path for the tailed file.

Also simplifies the logic of when to append the hash.

### DIFF
--- a/tail/tail_test.go
+++ b/tail/tail_test.go
@@ -320,9 +320,8 @@ func TestGetFileStateWithHashPathEnabled(t *testing.T) {
 		},
 	}
 
-	filename := "foobar.log"
-	osTempStatefilename := fmt.Sprintf("foobar.leash.state-%x", sha1.Sum([]byte(os.TempDir())))
-	tsTempStatefilename := fmt.Sprintf("foobar.leash.state-%x", sha1.Sum([]byte(ts.tmpdir)))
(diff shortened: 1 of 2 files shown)


robbkidd (commented):
thumbs up

❓ I might have missed it: do we have a test that two log files with the same basename and different paths result in state file names that do not match?

tail/tail_test.go
```diff
@@ -352,6 +351,22 @@ func TestGetFileStateWithHashPathEnabled(t *testing.T) {
...
+			ReadFrom:              "start",
+			Stop:                  true,
+			HashStateFileDirPaths: true,
+		},
+	}
+
+	statefile1 := getStateFile(conf, "/var/logs/app-1/foobar.log", 1)
+	statefile2 := getStateFile(conf, "/var/logs/app-2/foobar.log", 1)
+	if statefile1 == statefile2 {
+		t.Error("state files with different paths should not be equel")
```
robbkidd, line 367:
```suggestion
		t.Error("state files with different paths should not be equal")
```

robbkidd (approved):
👍 :shipit: 🚀 🍰

robbkidd (approved):

PR #199, opened 2021-05-11
Support darwin/arm64 (M1) & windows/amd64 (Win10)

lizthegrey:


robbkidd (approved):
The build changes make sense to me. 👍🏻

robbkidd:
It would be splendiferous if we could sort out some smoke testing in CI of behavior on non-Linux OSes.

PR #200, opened 2021-05-14
Update Go to 1.16 & test cross-compilation in PRs

robbkidd:
* test and build with Go 1.16 

Not only is it the latest and greatest, it allows for darwin-arm64 builds.

* matrixify the 'go build' for os/arch 

Turn the parameterized 'go-build' command into a parameterized 'go-build' job. As a job separate from the packaging, this can be added to the 'build' workflow as a job to perform on pull requests and not only run on git taggings.

For example, on this PR ...

a CI job flow diagram showing a "test" job as a dependency of four parallel "go-build-architecture-os" jobs

* remove Windows from the build matrix

The amd64 version throws errors during the build:

    # github.com/honeycombio/honeytail/tail
    tail/tail.go:321:13: undefined: unix.Stat_t
    tail/tail.go:322:12: undefined: unix.Stat
    tail/tail.go:431:13: undefined: unix.Stat_t
    tail/tail.go:432:2: undefined: unix.Stat
    Exited with code exit status 2

Probably need to use os.Stat() instead?

Also, need to figure out how to exclude the Windows + arm64 combination from the matrix. If we can't get 'exclude:' to work here, we can [add a conditional to the job](https://circleci.com/docs/2.0/reusing-config/#defining-conditional-steps) to do nothing.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,11 +7,11 @@ orbs:
 platform_matrix: &platform_matrix
   matrix:
     parameters:
-      os: &oses ["linux", "darwin", "windows"]
+      os: &oses ["linux", "darwin"] # <-- "windows" in the future!
       arch: &arches ["amd64", "arm64"]
-    exclude:
-      - os: "windows"
-      - arch: "arm64"
+    # exclude: # And when Windows comes, we'll need to exclude the Win+arm64 combo somehow
+    #   - os: "windows"
+    #   - arch: "arm64"
 
 jobs:
   test:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -41,12 +41,12 @@ jobs:
           GOOS=<< parameters.os >> \
           GOARCH=<< parameters.arch >> \
           go build -ldflags "-X main.BuildID=${CIRCLE_TAG}" \
-          -o $GOPATH/bin/honeytail-<< parameters.os >>-<< parameters.arch >> \
+          -o ~/artifacts/honeytail-<< parameters.os >>-<< parameters.arch >> \
           .
       - persist_to_workspace:
           root: ~/
           paths:
-            - bin
+            - artifacts
 
   build_packages:
     docker:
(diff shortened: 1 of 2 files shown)


build-pkg.sh
```diff
@@ -36,7 +36,7 @@ fpm -s dir -n honeytail \
     -t $pkg_type \
     -a $arch \
     --pre-install=./preinstall \
-    $GOPATH/bin/honeytail-linux-${arch}=/usr/bin/honeytail \
+    ./artifacts/honeytail-linux-${arch}=/usr/bin/honeytail \
```
vreynolds, line 39: We only use the build script in CI, right? Changing the directory to artifacts will make it break locally, but I don't know if we care, unless we build the packages locally.

.circleci/config.yml
```diff
@@ -4,65 +4,62 @@ orbs:
...
           .
-
-jobs:
-  test:
-    docker:
-      - image: circleci/golang:1.13
-    steps:
-      - checkout
-      - run: go test --timeout 10s -v ./...
+      - persist_to_workspace:
```
vreynolds, line 46: huh, are we expecting to see artifacts here: https://app.circleci.com/pipelines/github/honeycombio/honeytail/272/workflows/90c58c3d-4e1a-4d9c-8471-0a7132431256/jobs/843/artifacts

build-pkg.sh
robbkidd: It's my understanding that CI is the only thing using this script. Certainly it's the only automation in this repo that uses it. The original $GOPATH/bin path is also in test.sh, but CI doesn't appear to use that.

.circleci/config.yml
robbkidd: I don't think so. This is passing some file system context between jobs during a workflow. To have things appear in artifacts, that's the `store_artifacts` command (used in our `build_packages`)

vreynolds (approved):

.circleci/config.yml
robbkidd: I ... have yet to confirm that multiple concurrent jobs will be able to dump all their things in the same directory for `persist_to_workspace`. 😬  I aim to experiment with a `vx.y.z-rc1` tag to see them all come back together for the `build_packages` job after merge. 🤞🏻

robbkidd:
Let's do this!

build-pkg.sh
robbkidd: Update: I was wrong about `./artifacts`. All the jobs putting things in `~/artifacts` ... the default working directory for a job is `~/project`. So, yea, this broke.

PR #226, opened 2021-10-26
Parse trace data from SQL comments

endor:
If we want to associate the postgreSQL log data with existing traces/parent spans, we need some way to tell which trace/parent span the log data belongs to. One way to do this is to send the trace data as part of the query inside a SQL comment, similar to what https://google.github.io/sqlcommenter/ does. This trace data is then parsed here and send on to Honeycomb.


robbkidd (approved):
Thank you for the submission, @endor!

This looks like a good start to getting DB spans emitted from honeytail to participate in a trace. I foresee as a first step towards a future where honeytail parses the whole `/* … */` sql comment according to the sqlcommenter spec to return, say, a string map with which any number of fields can be added to the event.

Issue #250, opened 2022-06-20
postgres parser reports `duration` not `duration_ms`


pckilgore:
Honeytail  postgres parser sends `duration` field for events, not `duration_ms`.

**Versions**

- Go: go1.18.3
- Honeytail: 1.16.2

**Steps to reproduce**

1. Instrument postgres logs via honeytail
2. See `duration` reported in event, [which does not match the documented specification to send `duration_ms`](https://docs.honeycomb.io/getting-data-in/tracing/troubleshooting/#honeycomb-is-not-recognizing-my-traces)

Example event from console:
```json
{
  "duration": 0.085,
  "env": "test",
  "normalized_query": "SNIP",
  "pid": 19675,
  "query": "SNIP",
  "service.name": "SELECT snip FROM snip"
}
```

PR #251, opened 2022-06-20
fix(postgres): report query duration as duration_ms

pckilgore:
Per company documentation for auto-recognition for traces: https://docs.honeycomb.io/getting-data-in/tracing/troubleshooting/#honeycomb-is-not-recognizing-my-traces

- #250

- Reports duration under `duration_ms` vs. `duration`

### DIFF
--- a/parsers/postgresql/postgresql.go
+++ b/parsers/postgresql/postgresql.go
@@ -209,7 +209,7 @@ func (p *Parser) handleEvent(rawEvent []string) *event.Event {
 
 	if rawDuration, ok := slowQueryMeta["duration"]; ok {
 		duration, _ := strconv.ParseFloat(rawDuration, 64)
-		ev.Data["duration"] = duration
+		ev.Data["duration_ms"] = duration
 	} else {
 		logrus.WithField("query", query).Debug("Failed to find query duration in log line")
 	}


parsers/postgresql/postgresql.go
```diff
@@ -209,7 +209,7 @@ func (p *Parser) handleEvent(rawEvent []string) *event.Event {
 
 	if rawDuration, ok := slowQueryMeta["duration"]; ok {
 		duration, _ := strconv.ParseFloat(rawDuration, 64)
-		ev.Data["duration"] = duration
+		ev.Data["duration_ms"] = duration
```
robbkidd, line 212: I'm fully 👍🏻 with _adding_ a `duration_ms` field to honeytail events, but we shouldn't remove `duration` immediately.

robbkidd (changes requested):
Because field names are an implied schema for the backend receiving telemetry, we can't rename fields lightly. This will break any queries, triggers, or SLOs for current users of honeytail who care about event duration.

Adding new, better names is great, though! Can you update this PR to keep the old `duration` field and still add the new `duration_ms`?

parsers/postgresql/postgresql.go
pckilgore: [afd2ba8](https://github.com/honeycombio/honeytail/pull/251/commits/afd2ba8fd5f8c4ad57c3c8bf2233de5dfae24a19)

robbkidd (approved):
tenor-152337295

PR #262, opened 2022-08-24
Support YAML configs

### LINKED ISSUE #197: Honeytail 1.4.1 only supports a single value for AddFields in .conf
I haven't had a chance to repro this from source or bisect, but in upgrading from an ancient version of Honeytail to `1.4.1`, I noticed that the extra fields I'd been adding per-honeytail instance via `AddFields key=val` in the config were getting dropped.

Previously I was invoking Honeytail something like this:
`/usr/bin/honeytail -c /etc/honeytail/honeytail_nginx.conf`

But as a work-around I'm now doing this:
`/usr/bin/honeytail -c /etc/honeytail/honeytail_nginx.conf --add_field parser=nginx --add_field hostname=api-0.env.hostname.com --add_field service=api` etc.

This is mostly 6-of-1, half-dozen of another, but since AddFields still shows up in the default config, I was wondering if there was a regression (or if I'm simply not specifying it correctly).

Old configs looked like:
```AddFields = parser=nginx
AddFields = parser=nginx
AddFields = hostname=api-0.env.hostname.com
AddFields = service=api
```

Thanks!

### LINKED ISSUE #261: It's not possible to specify multi-valued fields (slices) in a .ini file
**Steps to reproduce**

1. Create an ini file with multiple values for AddFields (or other multi-valued fields)
2. Specify that ini file with --config; only the first value will be returned.

**Additional context**

The bug is that the jessevdk/go-flags package doesn't handle multiple entries for slice-valued items by default. You can, however, specify in the tags for the config object a delimiter like this: `env-delim:","`, which will split the field on that delimiter. The fix is to add that tag specification to all slice fields in main.go.

I tested this with a dummy app and it works.

Internal slack conversation: https://houndsh.slack.com/archives/C012HRW16HZ/p1660257971833609

kentquirk:

- Fixes #261 
- Closes #197 (we will not fix the .INI file problem, but deprecate .INI files instead) 

It turns out that the [library we use for parsing command lines and .INI configuration files](https://github.com/jessevdk/go-flags) has a bug where it just doesn't properly support multi-valued fields in INI files. The bug has existed for a while and the repository is stale. 

I *could* have forked the library, fixed the bug, and then we'd have to maintain a fork forever since it seems unlikely to ever be fixed upstream. 

Or I could have tried to switch to some other argument parser, but that would almost certainly have a whole collection of backwards-compatibility issues.

In addition, ini files are also pretty stale as a general concept. It seemed worthwhile to support a more modern format, and YAML is certainly eating the world as a config file format.

So I've done this:
- Chosen a popular yaml parser
- Decorated all of the configuration options with appropriate YAML tags to make sure that the configuration names are preserved
- Added a flag to write out a YAML config after reading the existing config (so that transitioning to the new format is easy)
- Added a flag to load a YAML config

Note that in the INI file world, because the config parser and the command line parser were the same entity, individual values in the config file could be overridden on the command line. But if you load a YAML file, all other config options are ignored.

No .INI files have been harmed in the making of this PR. All existing .INI behavior has been preserved -- this just lets us deprecate them and move to YAML instead.

The [Honeycomb documentation](https://docs.honeycomb.io/getting-data-in/logs/honeytail/) also needs to be updated. I'll put up a PR for that soon.

### DIFF
--- a/parsers/csv/csv.go
+++ b/parsers/csv/csv.go
@@ -17,10 +17,11 @@ import (
 // Options defines the options relevant to the CSV parser
 type Options struct {
 	Fields           string `long:"fields" description:"Comma separated list of CSV fields, in order."`
-	TimeFieldName    string `long:"timefield" description:"Name of the field that contains a timestamp"`
-	TimeFieldFormat  string `long:"time_format" description:"Timestamp format to use (strftime and Golang time.Parse supported)"`
-	NumParsers       int    `hidden:"true" description:"number of csv parsers to spin up"`
-	TrimLeadingSpace bool   `bool:"trim_leading_space" description:"trim leading whitespace in CSV fields and values" default:"false"`
+	TimeFieldName    string `long:"timefield" description:"Name of the field that contains a timestamp" yaml:"timefield,omitempty"`
+	TimeFieldFormat  string `long:"time_format" description:"Timestamp format to use (strftime and Golang time.Parse supported)" yaml:"time_format,omitempty"`
+	TrimLeadingSpace bool   `long:"trim_leading_space" description:"trim leading whitespace in CSV fields and values" yaml:"trim_leading_space,omitempty"`
+
+	NumParsers int `hidden:"true" description:"number of csv parsers to spin up" yaml:"-"`
 }
 
 // Parser implements the Parser interface
--- a/go.mod
+++ b/go.mod
@@ -18,6 +18,7 @@ require (
 	github.com/sirupsen/logrus v1.8.1
 	github.com/stretchr/testify v1.8.0
 	golang.org/x/sys v0.0.0-20210320140829-1e4c9ba3b0c4
+	gopkg.in/yaml.v3 v3.0.1
 )
 
 require (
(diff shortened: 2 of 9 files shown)


porterctrlz:
Bee still my beating heart :)

parsers/csv/csv.go
```diff
@@ -17,10 +17,11 @@ import (
 // Options defines the options relevant to the CSV parser
 type Options struct {
 	Fields           string `long:"fields" description:"Comma separated list of CSV fields, in order."`
```
MikeGoldsmith, line 19: Should we add a `yam:"-"` for Fields too?

MikeGoldsmith (approved):
Looks good to me.

I've left one suggestion to add an empty yaml declaration for Fields in the csv options struct.

parsers/csv/csv.go
kentquirk: Definitely not "-" -- that would prevent the field from ever showing up. The rule I used was to add a YAML declaration only when the field name used in config didn't match the variable name, or when it should be suppressed. In this case, the field is called "Fields" and that's also the config value name, so a YAML tag naming the field would be redundant.

PR #263, opened 2022-08-25
Don't delete a time field if we can't parse it as a time.

### LINKED ISSUE #238: GetTimestamp says it deletes the time value it returns, but not if it's already a time.Time
Documentation for GetTimestamp() says that it deletes the timestamp that it finds from the map that's passed in.

However, if this timestamp is already in time.Time format, it (simply returns)[https://github.com/honeycombio/honeytail/blob/0ff79e56031cc7281606b51e781da4e956a5abe7/httime/httime.go#L108].

There should be a delete before this return.

This was found by code inspection.

### LINKED ISSUE #239: GetTimestamp might delete a field that isn't a timestamp
In the [loop at the bottom of GetTimestamp()](https://github.com/honeycombio/honeytail/blob/0ff79e56031cc7281606b51e781da4e956a5abe7/httime/httime.go#L108), the function records a name to delete, and then attempts to parse its value. The result of that parse will be a zero Time value if the parse failed (and if so, the loop continues). But if the loop runs without ever finding a valid timestamp, it will still delete the last field that matched [one of the names](https://github.com/honeycombio/honeytail/blob/0ff79e56031cc7281606b51e781da4e956a5abe7/httime/httime.go#L38) it's searching. 

To be fair, if someone has called a field "Timestamp" and this code can't parse it, there are probably bigger problems. But I noticed this while reviewing the code.

kentquirk:

- Fixes #238
- Fixes #239 

- GetTimestamp violates its documented behavior (fails to delete the timestamp field) if a timestamp matches a certain format. This causes it to do that.
- GetTimestamp looks at a bunch of fields and tries to parse them as a timestamp. If it succeeds, it deletes the field. But there was a bug (see the referenced bug report). This fixes that bug.

### DIFF
--- a/httime/httime.go
+++ b/httime/httime.go
@@ -152,16 +152,18 @@ func GetTimestamp(m map[string]interface{}, timeFieldName, timeFieldFormat strin
 		delete(m, timeFieldName)
 		return ts
 	}
+
 	// go through all the possible fields that might have a timestamp
 	// for the first one we find, if it's a string field, try and parse it
-	// if we succeed, stop looking. Otherwise keep trying
+	// if we succeed, stop looking. Otherwise keep trying.
 	for _, timeField := range possibleTimeFieldNames {
 		if t, found := m[timeField]; found {
 			timeStr, found := t.(string)
 			if found {
-				foundFieldName = timeField
 				ts = tryTimeFormats(timeStr, timeFieldFormat)
 				if !ts.IsZero() {
+					// Only set foundFieldName if we were able to parse it as a time
+					foundFieldName = timeField
 					break
 				}
 				warnAboutTime(timeField, t, timeFoundInvalidFormatMsg)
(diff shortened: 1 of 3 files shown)


httime/httime.go
```diff
@@ -152,16 +152,18 @@ func GetTimestamp(m map[string]interface{}, timeFieldName, timeFieldFormat strin
...
+	// if we succeed, stop looking. Otherwise keep trying.
 	for _, timeField := range possibleTimeFieldNames {
 		if t, found := m[timeField]; found {
 			timeStr, found := t.(string)
 			if found {
-				foundFieldName = timeField
 				ts = tryTimeFormats(timeStr, timeFieldFormat)
 				if !ts.IsZero() {
+					// Only set foundFieldName if we were able to parse it as a time
+					foundFieldName = timeField
```
pkanal, line 166: Could we add a test for this if possible?

kentquirk: Good call. Added.

MikeGoldsmith (approved):
seal-of-approvement

PR #264, opened 2022-08-25
Remove dependency on mongodbtools

### LINKED ISSUE #257: Remove dependency on mongodbtools
[mongodbtools](https://github.com/honeycombio/mongodbtools) does not support newer log format and we'd like to sunset it.

Remove the dependency on `mongodbtools` and bring the code into `honeytail`. See if we can fix the log format support while we're at it, but as a stop gap measure, we can just file it as a follow-up issue.

kentquirk:

- Fixes #257
- Closes #258 
- Closes #259 
- Closes #260 

- mongodbtools is being sunset; import the appropriate bit of that library into this one. No functionality changes.


MikeGoldsmith (approved):
approved

PR #297, opened 2023-04-07
maint: Include LICENSES in distributions

TylerHelmuth:

- Includes our dependency's licenses in our distribution

- add makefile targets to update and verify licesnes
- add LICENSES folder with all dependency licenses
- Add job to CI to ensure LICENSES is up-to-date.


kentquirk (approved):
Ouch. This thing apparently thinks the Go mysql driver license requires full source distribution. I'm not sure I agree, but it's not worth arguing about.

TylerHelmuth:
@kentquirk ya based on the license type it makes decisions about whether or not the code requires only the licenses to be distributed or the source code as well.

PR #310, opened 2023-06-23
chore: use temp credentials for CI

NLincoln:

- Use temp credentials for CI

- Bump the orb version
- Use the role name instead of access keys

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -101,7 +101,7 @@ jobs:
     steps:
       - attach_workspace:
           at: ~/
-      - aws-cli/configure:
+      - aws-cli/setup:
           role-arn: "arn:aws:iam::702835727665:role/circleci-public-repos"
           role-session-name: "honeytail"
           aws-region: AWS_REGION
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -101,10 +101,9 @@ jobs:
     steps:
       - attach_workspace:
           at: ~/
-      - aws-cli/install
```
TylerHelmuth, line 104: Why does changing the credentials make this safe to remove?

TylerHelmuth (approved):

.circleci/config.yml
```diff
@@ -101,10 +101,9 @@ jobs:
...
       - aws-cli/configure:
-          aws-access-key-id: AWS_ACCESS_KEY_ID
-          aws-secret-access-key: AWS_SECRET_ACCESS_KEY
+          role-arn: "arn:aws:iam::702835727665:role/circleci-public-repos"
```
TylerHelmuth, line 105: Job is failing with `Cannot find a definition for command named aws-cli/configure`

PR #332, opened 2024-01-29
feat: Add config option for log level

### LINKED ISSUE #330: allow disabling logs
**Is your feature request related to a problem? Please describe.**
We use honeytail to forward logs to honeycomb from our ECS containers. In the ECS logs (which I have to look at sometimes as not all logs are forwarded), we keep seeing a ton of log lines caused by [this line.](https://github.com/honeycombio/honeytail/blob/f5861dc5d9cdc3a83e24f7b67140fd04523a08f4/response_stats.go#L85-L94) These logs pollute the log stream and make it harder to find other relevant logs outside of the Honeycomb interface.

**Describe the solution you'd like**
I'd love to be able to set an env var (i.e. `HONEYTAIL_NO_LOG`) that would prevent these logs from being written at all.

**Describe alternatives you've considered**
I've contemplated not using honeytail at all for logs, but migration away from the status quo is just not something we have capacity for at the moment.

**Additional context**
n/a

MikeGoldsmith:
Honeytail logs a summary of every batch of events sent to Honeycomb. These are very repetitive and obscure other information if there was a problem sending events. The summaries are logged at the `INFO` level using logrus.

This PR adds a log level configuration option that allows users to set their desired log level in the logging framework logrus. The usage of info level was verified to not contain critical information if there was a problem, with more important logs using a higher log level (eg `error`).

- Closes #330 

- Add new `log_level` option, defaulting to `info`
- If the new option has a value, attempt to set that log level in logrus

This was tested by setting the `log_level` flag after building the tool locally. As we use flag parsing in main file, it's not easy to write unit tests.

### DIFF
--- a/main.go
+++ b/main.go
@@ -92,6 +92,8 @@ type GlobalOptions struct {
 	FilterFiles         []string `short:"F" long:"filter-file" description:"Log file(s) to exclude from --file glob. May have multiple values, including multiple globs." yaml:"filter-file,omitempty"`
 	RenameFields        []string `long:"rename_field" description:"Format: 'before=after'. Rename field called 'before' from parsed lines to field name 'after' in Honeycomb events. May have multiple values." yaml:"rename_field,omitempty"`
 
+	LogLevel string `long:"log_level" description:"Set the log level. Valid values are 'debug', 'info', 'warn', 'error', 'fatal', 'panic'." default:"info" yaml:"log_level,omitempty"`
+
 	Reqs  RequiredOptions `group:"Required Options" yaml:"required_options,omitempty"`
 	Modes OtherModes      `group:"Other Modes" yaml:"-"`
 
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -92,6 +92,8 @@ type GlobalOptions struct {
 	FilterFiles         []string `short:"F" long:"filter-file" description:"Log file(s) to exclude from --file glob. May have multiple values, including multiple globs." yaml:"filter-file,omitempty"`
 	RenameFields        []string `long:"rename_field" description:"Format: 'before=after'. Rename field called 'before' from parsed lines to field name 'after' in Honeycomb events. May have multiple values." yaml:"rename_field,omitempty"`
 
+	LogLevel string `long:"log_level" description:"Set the log level. Valid values are 'debug', 'info', 'warn', 'error', 'fatal', 'panic'." default:"info" yaml:"log_level,omitempty"`
```
VinozzZ, line 95: Wow, I didn't know logrus supported `fatal` and `panic` levels. I wonder if users would find them useful when using honeytail

MikeGoldsmith: Yeah, probably not. I let Co-Pilot auto generate the comment here 😄 

We have log entries for debug, info, error and fatal. We could remove warn and panic as they are unused.

MikeGoldsmith: Coming back to this, I think the comment is valid - panic and warn log levels are valid, we just don't use them.

VinozzZ (approved):

Issue #359, opened 2025-08-07
Invalid systemd service file on arm DEB installs


jerm:
tl;dr the honeytail.service file that ships in the aarch64 DEB (at least) is broken. It'll let honeytail start, but not be enabled to start at boot time

Tested on Ubuntu aarch64 20.04, 22.04, 24.04. 
It may be broken elsewhere, but I'm not running any amd64/x86_64 boxes

**Versions**

- Honeytail: 1.8, 1.10, maybe others?

**Steps to reproduce**

1. install honeytail deb
2. try to enable honeytail in systemctl

`# systemctl enable honeytail`
> Failed to enable unit: Cannot alias honeytail.service as honeytail.

removing the alias line..
`# systemctl daemon-reload`
`# systemctl enable honeytail`

>The unit files have no installation config (WantedBy=, RequiredBy=, UpheldBy=,
Also=, or Alias= settings in the [Install] section, and DefaultInstance= for
template units). This means they are not meant to be enabled or disabled using systemctl.

adding `WantedBy=multi-user.target`
`# systemctl daemon-reload`
`# systemctl enable honeytail`
>Created symlink /etc/systemd/system/multi-user.target.wants/honeytail.service → /usr/lib/systemd/system/honeytail.service.

Lovely, and a reboot confirms that it starts as expected.

### Summary: 
Swapping-out

```
[Install]
Alias=honeytail honeytail.service
```

and replacing with

```
[Install]
WantedBy=multi-user.target
```

Allows honeytail to be enabled in systemctl

jerm:
@robbkidd just merged the fix. Thanks!! 😃

PR #360, opened 2025-08-07
Fix honeytail.service file so that honeytail can be enabled in systemd

jerm:

(Started with an issue (https://github.com/honeycombio/honeytail/issues/359) and realized it was more than just on DEB systems, and that the file was right here)

1) There's no need (anymore?) to alias honeytail to honeytail.service

2) as the file it now,It'll let honeytail start, but not be enabled to start at boot time

try to enable honeytail in systemctl

Failed to enable unit: Cannot alias honeytail.service as honeytail.

removing the alias line..

>The unit files have no installation config (WantedBy=, RequiredBy=, UpheldBy=, Also=, or Alias= settings in the [Install] section, and DefaultInstance= for template units). This means they are not meant to be enabled or disabled using systemctl.

adding `WantedBy=multi-user.target`

> Created symlink /etc/systemd/system/multi-user.target.wants/honeytail.service → /usr/lib/systemd/system/honeytail.service.

Lovely, and a reboot confirms that it starts as expected.

### DIFF
--- a/honeytail.service
+++ b/honeytail.service
@@ -12,4 +12,4 @@ RestartSec=5
 StartLimitInterval=0
 
 [Install]
-Alias=honeytail honeytail.service
+WantedBy=multi-user.target


robbkidd (approved):
Hunh. How did this ever work?

Yes! This is more better! Thank you!

robbkidd:
Thank you, @jerm! This fix is in today's [v1.11.0 release](https://github.com/honeycombio/honeytail/releases/tag/v1.11.0).

PR #361, opened 2026-03-20
chore: update to go 1.24, followup tidiness

robbkidd:

We were shipping builds based on Go 1.20 with dependencies over a year old. (`go.mod` said to use 1.21, but CI was building with 1.20.)

### Update Go to 1.24

* Update all dependencies to latest versions compatible with Go 1.24
  * Update `LICENSES`
* Fix `fmt.Fprintf` vet error
* Replace use of `rand.Seed()` (now a no-op since Go 1.22) in sampling tests by injecting a deterministic `*rand.Rand`
  * … which forced changes in the runtime code to plumb `*rand.Rand` in a few places to be injectable.
* Bump CircleCI images to Go 1.24 so we're testing and building the distributables with a consistent, recent Go

### CI Housekeeping

* Move the "how" of testing and building from CI config to `test` and `build` make targets
  * … which leaves CI with the "what" of calling those make targets
* Simplify (slightly) the `build_packages` job to run on `cimg/ruby` so it doesn't need to install so much stuff
* Modernize the Dockerfile to be a multi-stage build using more up-to-date base images


MikeGoldsmith (approved):

leash_test.go
```diff
@@ -453,19 +452,18 @@ func TestSampleRate(t *testing.T) {
...
 	opts.SampleRate = 3
 	opts.TailSample = true
-	run(context.Background(), opts)
-	// setting a sample rate of 3 gets 17 requests.
-	// tail does the sampling
+	rng := rand.New(rand.NewSource(1))
+	run(context.Background(), opts, rng)
+	// tail does the sampling with a deterministic RNG (seed 1, rate 3 → 17 of 50)
 	assert.Equal(t, ts.rsp.evtCounter, 17)
-	assert.Contains(t, ts.rsp.reqBody, `{"format":"json49"},"samplerate":3,`)
```
robbkidd, line 468: I should try to keep this assertion about recorded sample rate.

robbkidd: Added the assertion back in. The same expectation works with the new injectible deterministic RNG, too.

